PR #2, opened 2019-11-15
Add RFC: Built-in Conference Calling

VodBox:
### Description

Implement an over the web conference calling system that can be joined through a web URL, with separate video and audio feeds for each participant in OBS.

### Motivation and Context

Having an open source and peer to peer system that can directly ingest into OBS from a remote browser would allow for much easier collaboration and remote workflows in OBS. More details in the RFC.

### [View RFC](https://github.com/VodBox/rfcs/blob/conference-calling/accepted/0001-conference-calling.md)

### DIFF
--- a/accepted/0001-conference-calling.md
+++ b/accepted/0001-conference-calling.md
@@ -1,0 +1,40 @@
+- Start Date: 2019-11-15
+- RFC PR: #2
+- Mantis Issue: N/A
+
+# Summary
+
+Implement an over the web conference calling system that can be joined through a web URL, with separate video and audio feeds for each participant in OBS.
+
+# Motivation
+
+Currently the only option for easily bringing in separate video and audio feeds of a conference call into OBS is with Skype outputting NDI, and being picked up by the obs-ndi plugin. This solution requires the user to install a plugin that doesn't ship with OBS, and to use Skype which puts a watermark over the NDI feed. The solution is also almost entirely closed source.
+
+Having an open source and peer to peer system that can directly ingest into OBS would allow for much easier collaboration and remote workflows in OBS.
+
+# Detailed design
+
+The calling should be done over WebRTC, so that anyone with a modern web browser will be able to simply receive a link in order to join a call. As WebRTC is a big binary on it's own, the system in OBS should utilize a compatible but lightweight library. Since we'll want to be dealing with data directly, this can be achieved with either [librtcdc](https://github.com/xhs/librtcdc) or [librtcdcpp](https://github.com/chadnickbok/librtcdcpp), which implement only the data channel parts of the WebRTC standard.
+
+In Tools, there should be a menu item for creating calls. On clicking the menu item, a dialog should come up with an obsproject.com URL that can be copied to people who will participate in the call, an option to pick a webcam and mic to transmit back to the participants, and an option to mute the call mix going out to desktop audio. The OBS client will then communicate with a server running on obsproject.com to wait for clients to peer with it using WebRTC.
+
+When a participant opens the URL and grants permissions, it should communicate with the OBS client its capabilities (video, audio, screenshare, etc), and wait for the OBS client to request a media stream. When the client is requested to provide a stream, the webpage should start a MediaRecorder() instance and send the raw byte data over the WebRTC data channel.
+
+OBS will receive the video and audio over the WebRTC data channel, and send it to libavcodec to decode. It will then provide a video feed back over the data channel to all users that switches between cameras based on who is speaking. This will contain an audio mix of all participants.
+
+OBS should add a source for displaying one of these decoded feeds. This should contain a drop down to pick the source to pull into OBS, containing the following options.
+- Active Speaker, that will contain a mix of everyones audio, and will automatically switch video based on who is talking.
+- Active Remote Speaker, that does the same as above, but will not show the local feed.
+- Separate feeds for each of the participants in the call.
+
+The server hosted by OBS Project should also be open source. It will not handle broadcasting any of the video itself, merely facilitate the handshake between peers.
+
+# How We Teach This
+
+We will definitely need a guide written about using this system. A video will have to be made if this releases as part of an update. The source for call participants should have a label pointing people to the menu in tools to create a call, as a means of discoverability.
+
+# Drawbacks
+This requires OBS Project to host a server to facilitate P2P handshakes. Calling will be down if the host for the site is ever down, since a URL needs to be generated. Serving static HTML that can be cached for an entire directory may be a solution for this, if the server doesn't have to be active to handle the peering.
+
+# Alternatives
+Compiling OBS with the full WebRTC library and using the video channel instead. Another option is to provide a solution based around the existing browser source (although this may result in several more WebRTC connections to display feeds). Alternatively, we could find an existing open source conferencing program with a compatible license that could be integrated into OBS directly.


alinsavix:
So one thing I don't understand, is how connecting two people together is actually going to work. I understand they'd use an OBS server for coordination, but how do the users connect together? That ends up needing port forwarding or the like set up, since your average router/firewall only allows *outgoing* connections.

VodBox:
P2P connections can be formed without needing port forwarding on each end, so long as a server can help negotiate the connection (think BitTorrent and trackers). WebRTC lets you establish P2P connections between clients, and is in all modern web browsers.

kkartaltepe:
You can look up the most common implementation of NAT hole punching at https://en.wikipedia.org/wiki/STUN. In short, the relay server allows both ends to make outgoing connections to each other thus enabling the P2P connection in the presence of firewalls/NATs.

Lange:
This would also provide a lot of value for broadcasts which rely on rebroadcasting other RTMP streams.

For example, GDQ's Hotfix shows are "remote" shows wherein each speedrunner is actually streaming to a private RTMP ingest server, and our OBS then plays those RTMP streams. It's kind of a pain tbh.

Having a built-in system in OBS that did this better, with good controls, could be a more turnkey way of running these kinds of "remote" shows.

Palakis:
@Lange 
> This would also provide a lot of value for broadcasts which rely on rebroadcasting other RTMP streams.
> For example, GDQ's Hotfix shows are "remote" shows wherein each speedrunner is actually streaming to a private RTMP ingest server, and our OBS then plays those RTMP streams. It's kind of a pain tbh.

Would replacing RTMP with SRT be of any help in that situation?

Lange:
Hmm I don't think so. The protocol isn't the problem, it's the workflow. I don't think using SRT instead of RTMP would change the workflow, unless there is something about SRT I'm not aware of.

jp9000:
Would P2P be stable enough for this?

dodgepong:
I think P2P and STUN is sufficient and quite viable. I don't think we need to run a TURN server, as that would be too cost-prohibitive.

DDRBoxman:
Any concerns with librtcdcpp and librtcdc not getting updates in around the last 3 years?

Also how much work is involved building the video parts of webrtc on top of the data channels?

notr1ch:
The libraries suggested in the RFC are for using WebRTC data channels, this will likely not inter-operate with browser WebRTC which uses SRTP. These suggestions may need to be removed or replaced with more appropriate libraries.

I also have concerns about duplication of functionality as our browser source likely already has functional WebRTC support. Basic functionality like bringing in a single remote participant could possibly be handled through a new source type that sets up a pre-configured browser source.

VodBox:
WebRTC data channels are in all browsers supporting WebRTC (except iOS Safari), and are done over SCTP. Additionally, MediaRecorder is available in all desktop browsers except IE11 and Safari. The reason for selecting data channels is that we have full control over the data we send and
settings around retransmissions on both sides, and it means that we don't need a library to support the entirety of WebRTC, so we can pick more lightweight libraries. The libraries mentioned are just suggestions I had at the time, and there are people in the video space who have written
libraries for WebRTC who have expressed interest in helping out should work start. I also ran this RFC by a number of people working with WebRTC, who saw no immediate problems with my proposed design.

I mention in my alternatives section that we can use the full WebRTC lib used by browsers, or make use of the browser source. The problem with the former is that it's a heavyweight dependency to have, is duplicated from the browser we already have, and most of it would be unused. The problem
with the latter is how do we transmit video back to call participants, and in the case of many call participants on screen, each having their own browser source would result in a lot of needless overhead, on top of WebRTC not working on the macOS browser source entirely.

steveseguin:
Hi everyone.  So I'm the maintainer at https://github.com/steveseguin/obsninja, which is an open-source and free project designed to accomplish what is being talked about here. I'm doing it via a Browser Source and webRTC. 


Anyways, I like the vision being talked about and I figure I'd share some of my learnings re; WebRTC.  

Costs of anything server related are pretty substantial. TURN servers hosted by sites like Millicast or Twillio cost quite a bit. TURN servers though are often needed in WebRTC to support mobile devices, IPv6, and some firewalls. Many other users will still wish to use TURN servers as it enables them IP-privacy.

I've had some luck with a subset of technical users setting up their own TURN servers, but I've discovered that it is a requirement to offer TURN servers by default. I thought I could get around this, but just too many users require them. Most users do not understand why things fail and it was causing an overwhelming number of support requests.

I have a plan on how to ensure I can provide TURN servers for free to my users going into the future, but for the next several months I've decided to just pay out-of-pocket.

As for Group Conferencing, an SFU/MFU may also be required to scale a video conference past 4 members. Mesh networks are largely unreliable, and in the contenx of OBS.Ninja at least, they take up a lot of system resources. (CPU, GPU, network).  SFU/MFUs are quite expensive to host as well.

While I am offering a mesh-setup for small groups, for larger groups I am intending to offer integrations with hosted chat services. Regardless of SFU or Mesh, I would still need to offer TURN servers to ensure all group participants can join. 

As mentioned, MacOS Browser Source has become a pain point, as OBS v23 crashes for many users, and newer versions do not work with WebRTC at all. StreamLabs OBS does work, but it lacks hardware acceleration, and the video frame rate suffers.

I do prefer the "Browser Source" approach to solving this problem, as it makes the underlying code more accessible to more developers who may wish to change it up. For example, SDP-editing is desirable for me personally, as is the option of pulling out stats or remote controlling videos from a remote control panel. While OBS.Ninja does target OBS explicitly, it is also technically agnostic of platforms, and that increases its flexibility across the industry.

As a side note, I would love to work with OBSProject going forward in some fashion, and I also welcome contributions. I love OBS Studio and want to see it continue to blossom.

VodBox:
I feel as though we would only be able to offer TURN if it is provided to us free of charge by someone willing to sponsor it, or if we can figure out a way to make the service self-sustaining. At least for an initial revision of the RFC as described, I don't see TURN being offered in any way except for providing the server software for someone to self-run. This could easily change with time or circumstances.

While I understand that the browser source approach makes it more accessible to more developers, and it does have the nice benefit of needing one less dependency, it does still produce overhead, especially if we want to be able to bring in individual video feeds that can be manipulated as their own sources. Even if WebRTC in CEF on macOS could be resolved, I feel as though running a subset of the WebRTC stack natively, and decoding directly with libavcodec to a source would be enough of a save on resources to be worth it.

There will still be a need for developers who are familiar with frontend development with WebRTC for the browser clients, especially since control over retransmits, responding to buffer health, and how to balance quality vs. dropped frames, is all things that will need to be done with logic running on the remote client side. Indeed, a minimum viable product that doesn't transmit video back to a call participant would contain most of the actual logic happening on the remote client.

If there are concerns about control and availability of stats with a native solution, then we can always make sure that the native solution provides what you're looking for, or at the very least easily expose that information to developers. A potential solution would be to have a page on the OBS website that can connect to the running local OBS as a web control panel.

VodBox:
Just a follow up on viability of the design as proposed and working in the browser: [Mux recently put out an article](https://mux.com/blog/the-state-of-going-live-from-a-browser/) which shows the MediaRecorder API being used with WebSockets, and how that can be easily piped into FFmpeg directly. Zoom has also made their browser client [use WebRTC data channels for transmitting it's video](https://webrtchacks.com/zoom-avoids-using-webrtc/#datachannels), but falls back to WebSockets in place of STUN/TURN. The design is not something unproven, the issue is making it scale well, offer the control we want, and making it work in OBS.

Fenrirthviti:
This one has been open for a while, but was written after a lot of internal discussion, and the general concept of this and the idea of a remote caller is something we'd like to see, so I will enter this in to FCP with disposition merge.

PR #6, opened 2020-01-13
Add RFC: Media controls

cg2121:
### Description
Add RFC for a media control widget to control media and VLC sources.

### Motivation and Context
Users want more powerful control over their media sources.

### Link to RFC
https://github.com/cg2121/rfcs/blob/media-controls/accepted/0005-media-controls.md

### DIFF
--- a/accepted/0005-media-controls.md
+++ b/accepted/0005-media-controls.md
@@ -1,0 +1,49 @@
+- Start Date: 2020-01-13
+- RFC PR: #6
+- Mantis Issue: https://obsproject.com/mantis/view.php?id=329
+
+- Related PRs:
+	1. https://github.com/obsproject/obs-studio/pull/2300
+	2. https://github.com/obsproject/obs-studio/pull/2274
+	3. https://github.com/obsproject/obs-studio/pull/1803
+
+# Summary
+
+Create the ability for users to control the media and VLC sources.
+
+# Motivation
+
+Users should be able to have more control over media sources. With a media control widget, they can see how much time is left in the video, seek to a certain point, play/pause, etc.
+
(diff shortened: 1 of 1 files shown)


derrod:
1) Would this also add hotkeys or API functions to the media source for playback control (e.g. for a streamdeck)?

2) What happens with multiple media sources in a scene?

3) Some users want the ability to show clips on command, currently this is a bit hacky since it requires hiding/unhiding the source each time to reset. An option to not automatically start playback (and stay blank when a scene is added/becomes active) until the user manually triggers it would be desirable for that.

4) If a clip has finished, would clicking play again restart it? This would be required for 3) and match the behaviour of media players

cg2121:
1. Yes, it would add hotkeys.

2. In the previous PR (#1803) there were a widget for every media source in a scene, similar to the volume meters. I now find this would be a little overwhelming for users. My proposal now is have the widget update to the source when an user clicks on the source.

3. The previous PR added playback options for the media source. The options added were: restart on activation of source, pause when source becomes inactive and always play. These options are already in the VLC source.

4. Yes, the play/pause icon would change to a restart icon when playback finishes.

Fenrirthviti:
At this point, I feel that for this to have the most successful impact, it would rely on https://github.com/obsproject/obs-studio/pull/1838 for the UI/UX part of it. Since #1838 is a bit bigger in scope, I'm not sure the best way to bring these two together. A widget seems like it could work, but it would be better if we had a context menu that all plugins could put UI elements in, which would be easier and more accessible for everyone going forward.

Fenrirthviti:
This is also already complete, so merging.

PR #7, opened 2020-01-14
Add RFC: sRGB Color Space setting

jpark37:
EDIT (2022-12-22): Current thinking is that there should be no color conversion between sRGB and Rec. 709. So a lot of what is stated on this RFC might be backwards. In particular, the test matrix entries for Good and Correct should be flipped. In any case, OBS is (mostly) doing the right thing (most of the time).

### Description
Add sRGB "Color Space" to output options, and possibly change default to sRGB.

### Motivation and Context
There is often confusion around the "Color Space" option for output, and it has several problems.

- We don't actually convert colors between RGB color spaces in OBS. We merely tag the metadata.
- Rec. 601 output support is weird. x264 tags 601 as "undef" for everything, and jim-nvenc uses the European 625-line variant of 601 attributes.
- Any PC applications/games that output sRGB are tagged as 601/709, and will be handled incorrectly by color-accurate applications, e.g. Chrome on Mac.

Adding and defaulting to sRGB is probably the least of all evils.

- sRGB is identical to Rec. 709 except for the sRGB transfer function, which can be a free conversion for most GPUs. When represented as YCbCr, it makes sense to transform with Rec. 709 matrix coefficients.
- PC applications/games that use sRGB (very common) are now tagged accurately if using sRGB. Video capture feeds that use 601/709 would not be, but we can leave behind 601/709 settings for passthrough scenarios. Actually converting the color data is beyond the scope of this RFC.
- Chrome supports sRGB videos properly on all tested platforms.
- Rec. 601 for output was almost always the wrong choice.

### Test Matrix

| Platform | Application | 601 | 709 | sRGB (709 matrix) | Full | 480p |
| --- | --- | --- | --- | --- | --- | --- |
| Windows | Chrome | | Good (sRGB TRC) | Correct | Correct | |
| Windows | Firefox | | Good (sRGB TRC) | Correct | Bad | |
| Windows | VLC | | Good (sRGB TRC) | Correct | Bad | |
| Windows | Movies & TV | | Good (sRGB TRC) | Correct | | |
| Windows | Resolve | | Good (sRGB TRC) | Correct | | Bad (601 matrix) |
| Mac | Chrome | | Correct | Correct | Correct | | |
| Mac | Safari | | Correct | Correct | Correct | | |
| Android (Pixel XL) | Chrome | Good (709 prim.) | Bad (601 matrix) | Bad (601 matrix) | Bad | |
| Android (Pixel XL) | Twitch | Good (709 prim.) | Bad (601 matrix) | Bad (601 matrix) | Bad | |
| Android (Pixel XL) | VLC | | Bad (601 matrix) | Bad (601 matrix) | Bad | |
| Android (S9) | Chrome | Bad (709 matrix?) | Good (sRGB TRC) | Correct | Bad | |
| Android (S9) | Twitch | Bad (709 matrix?) | Good (sRGB TRC) | Correct | Bad | |
| iOS | Twitch | | Good (sRGB TRC) | Correct | Correct | |
| Twitch | (transcode) | | Correct (maybe) | Good (709 TRC) | Bad | |

### DIFF
--- a/accepted/0007-color-space-srgb.md
+++ b/accepted/0007-color-space-srgb.md
@@ -1,0 +1,33 @@
+# Summary
+
+Add sRGB to "Color Space" output options, and default to sRGB.
+
+# Motivation
+
+There is often confusion around the "Color Space" option for output, and it has several problems.
+
+- We don't actually convert colors between RGB color spaces in OBS. We merely tag the metadata.
+- Rec. 601 output support is weird. x264 tags 601 as "undef" for everything, and jim-nvenc uses the European 625-line variant of 601 attributes.
+- Any PC applications/games that output sRGB are tagged as 601/709, and will be handled incorrectly by color-accurate applications, e.g. old versions of Chrome handled 709 properly before they decided to cheat to save power.
+
+Defaulting to sRGB is probably the least of all evils.
+
+- sRGB is identical to Rec. 709 except for the sRGB transfer function, which can be a free conversion for most GPUs. When represented as YCbCr, it makes sense to transform with Rec. 709 matrix coefficients.
+- PC applications/games that use sRGB (very common) are now tagged accurately if using sRGB. Video capture feeds that use 601/709 would not be, but we can leave behind 601/709 settings for passthrough scenarios. Actually converting the color data is beyond the scope of this RFC.
+- Chrome supports sRGB videos properly on all tested platforms.
+- Rec. 601 for output was almost always the wrong choice.
+
+Implementation:
+
+- Add VIDEO_CS_SRGB, and plumb into all switch cases.
+- Go through output encoder settings, and use sRGB metadata values.
+
+# Drawbacks
+
+People that had a working color pipeline with the old faulty settings may have to tweak their setup.
+
+There is supposedly a performance regression involving DeckLink when switching away from 601.
+
+# Additional Information
+
+None.


Fenrirthviti:
Looks pretty good to me, only consideration I don't see addressed is maybe adding the option to change the format with a "hidden" .ini setting, for those ultra-niche use cases?

derrod:
> Rec. 601 for output was almost always the wrong choice. Was there ever a reason to support it?

I think the reason it is (was?) default was a recommendation from Twitch because of how browsers handled 709 incorrectly.

On the other hand some applications (e.g. VLC) always assume 601 for sub-HD resolutions and 709 for HD resolutions, which further complicates things. So if you are recording in an SD resolution for whatever reason, using 601 would likely lead to more accurate reproduction with players that make this assumption. Likewise recording in HD with the default setting would produce a worse result.

derrod:
Also while capturing stuff directly from a PC is surely the most common use case; there are a good amount of people simply using OBS as an encoder for a signal received from a traditional production setup (e.g. ESL and other broadcasters). Based on the description, wouldn't the change mean that their streams are then tagged incorrectly?

Generally compatability is probably something that needs to be cleared. There's more than just Chrome out there, thankfully.

jpark37:
I searched a bit, and couldn't find 601 as recommended by Twitch. I'm pretty sure I've seen it go haywire in Firefox from past testing, but I forget in what way. That might have been with real American 601 metadata settings though.

I'll put together a test video to compare 601/709/sRGB and 480p/1080p through Chrome, Firefox, VLC, and Davinci, and see what's up. I'll also try to see how Twitch handles sRGB transcodes.

As Fenrir mentioned, we could have a hidden ini setting for strict passthrough cases.

derrod:
To clarify: this wasn't a publicly documented recommendation by Twitch, but something they recommended directly in the early days of OBS as far as I know.

Not sure how much those would be affected but also to keep in mind:

 * Third party output plugins (e.g. NDI, VirtualCam)
 * DeckLink output

While I think changing the default and removing 601 is probably a good idea, I think 709 should at least remain easily selectable since streaming from an external source is a somewhat common use case. And digging through undocumented ini files doesn't sound like a good solution.

I have to admit I'm a bit out of my wheelhouse here, so forgive my ignorance, but seeing as sYCC seems to rather rarely discussed or documented (including only a total of three total mentions in the FFmpeg source code) I feel somewhat uneasy about removing existing options that are "fine" in favor of one that seems to be rather unexplored territory.

mufunyo:
Strongly disagree about removing the options. Advanced options are for advanced use, they should not be removed simply because unadvanced users can access them.

jpark37:
There are more search hits in FFmpeg for AVCOL_TRC_IEC61966_2_1. That gamma curve only applies to sRGB.

I'd agree about having this as an advanced setting if the options behaved as one would expect, but since they don't, e.g. 601+x264 = "undef" for all metadata values probably being the weirdest, I'd rather obscure them as "LegacyColorSpaceOutput" in an ini file, and not expose these weird behaviors in the UI.

Edit: In a future where specifying a color space actually converted color values, that may be more reasonable to have in the settings UI.

mufunyo:
If you expect WYSIWYG, then sure. But capturing a console or otherwise 709 source will show up in OBS as sRGB/sYCC incorrectly, and keeping the numbers but tagging the output as 709 should at least cause the output to be correct in players that honour the colour space flag properly.

coral:
Hey,

Coral from Twitch here. We're discussing this internally and will get back with what we think. I'm not suggesting that we want to mandate a decision but Twitch got brought up so I wanted to provide the context from our end.

jpark37:
For your reference, here is a Chromium issue that I filed a while ago about how they treat 709 videos as if they use the sRGB transfer function. It doesn't seem like they want to do anything about it.

https://bugs.chromium.org/p/chromium/issues/detail?id=947798

Edit: Chrome for Mac surprisingly makes the 709 adjustment though.

JohannMG:
I'm all for removing settings but is advanced settings a priority? 

> sYCC is identical to Rec. 709 except for the sRGB transfer function, which can be a free conversion for most GPUs. You can also think of sYCC as identical to sRGB, but transformed with Rec. 709 matrix coefficients; same difference.


Streaming Q: 
 - Will the transfer issue be a problem for some stream destinations expecting REC709 looking washed out? 
 - I know many people are using OBS for game streaming sRGB source but there's also people pulling in Rec709 cameras in: will those convert correctly? 

Recording destination Qs: 
- In Resolve will it transfer correctly:  I want to see what you see in Resolve Scopes. I don't think I've ever conformed in sYCC before. 
- Will people recording to file -> directly uploading to YouTube get the same results as the 709 tags previously?

mufunyo:
> * In Resolve will it transfer correctly:  I want to see what you see in Resolve Scopes. I don't think I've ever conformed in sYCC before.

Resolve operates in RGB, so this is called sRGB in Resolve. You can transfer to and from sRGB using the Color Space Transform OFX.

jpark37:
This particular setting is just badly implemented, and it tends to come up in Discord every so often. There's a lot of misunderstanding about how it works because how could you unless you've seen the source code.

The issue of what's currently correct and not is complicated by the most popular consumer of streaming content (Chrome, probably) treating 709 as sRGB. In the case of sRGB content (PC applications/games), current OBS claims sRGB as 709, and Chrome claims 709 as sRGB, so two wrongs cancel each other out, and you get correct footage. Footage from 709 sources like modern game consoles are actually not as bright as they should be (at least when using OBS (709/709) -> Twitch -> Chrome on PC).

Resolve and YouTube would need to be tested.

jpark37:
The above explanation is just for 709. The current state of 601 is weirder, and I don't want to type it all out at the moment.

JohannMG:
> so two wrongs cancel each other out, and you get correct footage

LOL okay I see. That's a mess. 

> 601 

okay but this RFC may benefit from it

@mufunyo My point is ideally the change simplifies things for users doesn't require to educate users to change color spaces in Resolve, or any NLE. And that this works with both screen recordings and camera/webcam footage. @jpark37 sounds like you have resolve covered but if you can send maybe example files I can compare in Premiere on as well?

mufunyo:
>  ideally the change simplifies things for users doesn't require to educate users to change color spaces in Resolve, or any NLE.

You need to abandon this idea. Colour management is completely and utterly broken on all platforms (including Mac), and as a content creator you can choose between blissful ignorance or education. In the latter case, you have to be aware yourself what colour space your inputs and outputs are, and create the proper monitoring conditions. Resolve on Windows can correctly display colours, but only if you know what you are doing, or if you accidentally create conditions in which the colours are somewhat accurate (think of the two wrongs cancelling each other out mentioned above). On Mac it's even worse and you can choose between Apple colour managed monitoring that is guaranteed to be wrong, or unmanaged, in which case Resolve behaves just like it does on Windows and you have to do your own manual colour management.

coral:
Ok can I ask a stupid question:

Does iOS, Android, FireTV etc parse sYCC natively?

I agree that the current setting causes confusion for users but there's a long tail of devices that needs to support sYCC if that's the replacement.

jpark37:
I have an Android phone (Pixel XL). If the phone video player is the same one in Chromium, then it should be fine in theory? I can test that device at least. I have an old iPad lying around somewhere, so I can test that as well.

The difference between 709 and sRGB comes down to a single H.264 metadata field. My possibly naive thought is that applications that are not written to handle the sRGB curve will likely fall back to whatever it would have done for 709, so for those applications, the behavior shouldn't get worse. I can at least test the most popular applications though.

JohannMG:
I'll also add that I saw a few times in Discord Rec. 2020 support was brought up: If there's any work on that in the future we'd have to undo all the sYCC defaults?

jpark37:
Well, if we're talking about having a buried ini setting, nothing would be hard-coded. Even if we decided to hard-code instead, it's not much work to make configurable again.

I don't want to talk too much about HDR here because it's a bit off-topic.

Xaymar:
What's the device support for sYCC playback? Rec.601 and Rec.709 have hardware support on many devices, and I'm not sure how sYCC compares to that given that it has a different transfer function. Especially the following devices and browsers interest me here:

* Android 7 and above. (GammaTransfer has Transfer_sYCC entry in qcom)
* iOS on iPhone 8 and newer.
* Windows 10 file playback using standard apps (Movies & TV on Win10)
* Mac file playback using standard apps
* Chrome
* Firefox
* Opera
* Safari

If the support for sYCC is poor, or causes higher battery usage, a better RFC would be to fix the color conversion to convert from one to the other color space instead of forcing an incompatible, higher battery usage color space. Battery life on mobile devices, which make up a large margin of viewer numbers, is very important and should be considered here.

Not only that, but we should also consider software and platform support here. Which platforms support sYCC decoding correctly? Can you upload sYCC content to YouTube and get the exact same result as you did with the fake-709 color space we had before or will it look massively different to before? How about editing software like Adobe Premiere Pro, Sony Vegas Pro, Windows Movie Maker (still being used today), DaVinci Resolve, Blender, Shotcut, OpenSHot, etc?

I'm against removing the option, but for changing the default to something that makes sense. And also for fixing all the color space issues that sources currently have that deal with a different color space.

Edit:

> I'm pretty sure I've seen it go haywire in Firefox from past testing, but I forget in what way.

Both Chrome and Firefox used to decode Rec.709 as Rec.601 causing colors to be very wrong. This has been fixed for over a year apparently.

jpark37:
I've started to do some testing. Here are a couple test images, and short videos recorded with them. I've verified with an external program that the metadata is set correctly.
[sYCC.zip](https://github.com/obsproject/rfcs/files/4063250/sYCC.zip)

I have uploaded the videos to YouTube.

709, full, 480p: https://www.youtube.com/watch?v=YZiAZ3wsHlw
709, full, 1080p: https://www.youtube.com/watch?v=sexKKuQo-40
709, limited, 480p: https://www.youtube.com/watch?v=2CXUmboDYBA
709, limited, 1080p: https://www.youtube.com/watch?v=fSCJWPOO7iA
sRGB, full, 480p: https://www.youtube.com/watch?v=JSNi8kThlxw
sRGB, full, 1080p: https://www.youtube.com/watch?v=QmTfE4w4Gog
sRGB, limited, 480p: https://www.youtube.com/watch?v=g84uTDZxUjE
sRGB, limited, 1080p: https://www.youtube.com/watch?v=f4mrOSuhKEU

Testing with latest Chrome and Firefox on Windows against YouTube, all sRGB videos played identically to the test images. All 709 videos were also identical even though they should be a little brighter since the colors were not adjusted for the different transfer functions. Regardless, this is fine. (Edit: Android Youtube app and Mac Safari also showed identical images between 709 and sRGB.)

Testing with VLC on Windows, limited range videos were correct with sRGB, but too dark with full range. 709 videos were identical to sRGB even though they shouldn't be. No regression here either.

Edit: Windows "Movies & TV" is a little wonky, but 709 == sRGB again at least.

I'm looking into testing Twitch and transcodes next.

jpark37:
Forgot I don't have transcodes on this account, but I can at least test the main video feed.

709, full, 480p: https://www.twitch.tv/videos/536276918
709, full, 1080p: https://www.twitch.tv/videos/536276978
709, limited, 480p: https://www.twitch.tv/videos/536277029
709, limited, 1080p: https://www.twitch.tv/videos/536277069
sRGB, full, 480p: https://www.twitch.tv/videos/536277105
sRGB, full, 1080p: https://www.twitch.tv/videos/536277136
sRGB, limited, 480p: https://www.twitch.tv/videos/536277169
sRGB, limited, 1080p: https://www.twitch.tv/videos/536277201

Mac Safari actually makes the brightness adjustment for 709 here.
Edit 2: Mac Chrome as well.
Windows Chrome does not, but is otherwise fine.
Windows Firefox is like Windows Chrome, but also fails to handle full range correctly for both sides.
Edit: Android Twitch app is like Windows Firefox although I'm not sure why the colors are off. The screenshot PNGs might not be sRGB, but whatever it is isn't sRGB specific.

No crazy regressions so far.

derrod:
Here the same samples from my account, which does have transcodes (I believe all of those should be using the QuickSync transcodes, that probably doesn't matter):

709, full, 480p: https://www.twitch.tv/videos/536287848
709, full, 1080p: https://www.twitch.tv/videos/536287627
709, limited, 480p: https://www.twitch.tv/videos/536287427
709, limited, 1080p: https://www.twitch.tv/videos/536288053
sYCC, full, 480p: https://www.twitch.tv/videos/536286522
sYCC, full, 1080p: https://www.twitch.tv/videos/536286260
sYCC, limited, 480p: https://www.twitch.tv/videos/536285901
sYCC, limited, 1080p: https://www.twitch.tv/videos/536285647

jpark37:
Looking at 180p transcodes with Mac Safari, full range is not handled properly for either color space. Moreover, sRGB is incorrectly displayed with what looks like the 709 correction where it was fine without the transcode. The likely cause is that the transfer function value is not being passed through to the new stream, but I don't really want to compile a web browser from source to verify that.

sRGB still appears to be viable. I'll probably want to look at iOS and some video editors, but it's looking less and less likely that something will fail catastrophically by making this simplification.

Xaymar:
I've tested the footage on my Sony Xperia XZ2 running Android 10. While I could not see any differences with a naked eye, a DSLR helped show them. sYCC and 709 looked identical, and there were differences to the sRGB images. The same can be said about DaVinci Resolve. After fixing the video range, both 709 and sYCC look identical, but the sRGB sample is visibly different again. 

Note that both of these tests were done after remuxing to mp4 as neither the android device or the software supports Matroska playback.

Edit: On a side note, there are even visible difference between limited and full range footage, which hints at the full range or limited range files not being properly scaled into their expected color range.

jpark37:
Using Resolve with MP4 variants, the 1080p videos seem to match the sRGB image, but the 480p videos are pretty far off. This might be a case where the application assumes 601 for SD resolutions rather than inspecting the metadata. sRGB seems to be doing well enough though.

jpark37:
Apparently, the Android Twitch app will assume 601 regardless of metadata or resolution. I was able to resolve the disparity seen with both 709 and sRGB by setting OBS to 601. This is the only use case for 601 video output I have found so far.

Xaymar:
> Using Resolve with MP4 variants, the 1080p videos seem to match the sRGB image, but the 480p videos are pretty far off. This might be a case where the application assumes 601 for SD resolutions rather than inspecting the metadata. sYCC seems to be doing well enough though.

sYCC and 709 are identical for me, but neither of them matches sRGB, they do get "close" though, within 1-2 rgb values of the correct color. I don't know enough about the standard to say that these differences are just something that is coming from the color space or not.

jpark37:
Unlike its Android counterpart, The Twitch app on iPad competently handles full range and 709/sRGB, only skipping the transfer function adjustment for 709.

coral:
OK from the brief discussions I've had internally, it seems most people agree that Twitch prefers 709/sYCC as a standard. 

I'll forward the 601 bug to the Android team.

jpark37:
Thanks. I've summarized the test results into a table up top. Let me know if anything I'm saying isn't clear.

coral:
@jpark37 What Android device did you use?

jpark37:
Pixel XL

coral:
@jpark37 If you have an extra minute, could you try a TS segment on VLC or some other player on Android and see if that respects it properly?

jpark37:
I can try that after work although I'm not familiar with HLS.

jpark37:
Chrome and VLC seem to be having similar issues on Android. Maybe this is a MediaPlayer issue?

coral:
@jpark37 It might be the hardware decoder on the Pixel XL not respecting the flags. Getting another android device with a different chip would help determine that. Probably outside the scope of this investigation however.

jpark37:
I have access to an S9, and it's reversed there; 709 looks good and 601 looks bad. I also just received a Pixel 4 XL, so I can try that tonight.

I'm also curious to see if I can get the GL surface type of the ExoPlayer demo app to work on my Pixel XL because I see the renderer has code to handle both 601 and 709 spaces.

coral:
@jpark37 Appreciate all the effort you're spending on this. If you figure out a way of convincing the Pixel into doing the right thing, let me know and I'll forward that to our Android team.

jpark37:
I can get ExoPlayer to respect 709 on the Pixel XL, and display the correct colors if I modify OBS to set metadata on the MKV container itself, and not just the video stream within. Otherwise, MatroskaExtractor ignores the color metadata completely. I'm also using the default SURFACE_TYPE_SURFACE_VIEW; SURFACE_TYPE_VIDEO_DECODER_GL_SURFACE_VIEW wasn't necessary (and may only work for VP9/AV1 at the moment).

The only change I needed to make to the ExoPlayer demo app was to replace one of the entries with my own local file.
Movie file goes here: demos/main/src/main/assets/
media.exolist.json uri: file:///android_asset/filename.mkv

Even if the Twitch app is using ExoPlayer, it probably isn't going through MatroskaExtractor, so the color info may need to be parsed and passed along through some other means, but I think it should be doable.

energizerfellow:
For NLE testing I'd check out Premiere, Resolve, Vegas, and Final Cut Pro.  Anecdotally that covers the majority of NLEs I've seen among OBS's user base, in roughly that order of popularity.

It also appears that Resolve uses BMD's in-house codecs in the paid version of Resolve, but the OS-side codecs in the free version, leading to (at least) notable performance variation between license levels. I don't have Resolve to test this, but I've seen a number of reports from the field that this is the case.  Testing with the paid license of Resolve is probably required here.

Making sure the color metadata is set in both the video data and the container should make them all happier, along with the audio keyframe fixes.

Anybody got any AppleTV, Roku, or Android TV (Nvidia Shield TV, Sony TV) devices kicking around for testing? LG webOS TV? Samsung Tizen TV?

Xaymar:
I have the paid version for Resolve (Resolve Studio) if needed for testing. And a Samsung Tizen TV.

mufunyo:
There is no functional difference between the paid and free versions of Resolve in this context. The paid version allows hardware decoding, but foregoing potential driver bugs the output should be identical to software decoding.

jpark37:
Updated RFC to be about adding the sRGB setting as an additional option that is default (currently 601), and not to remove the existing settings since people seem to want to keep them for passthrough scenarios.

mufunyo:
Twitch has given this their OK, YouTube seems fine, do I need to test Resolve still or is this good to go?

jpark37:
I'm okay with closing this out and trying to wrap up the associated sRGB PR, but I really hope we don't rearrange priorities because someone obnoxious and pushy appeared in Discord. Rewarding that behavior sets a terrible precedent.

Close out the issue no sooner than you would have if he hadn't come along please.

mufunyo:
To be perfectly clear, to me this is no different to sitting on the toilet one day thinking "hmm, that sYCC RFC, whatever happened to that again?". While I agree with the general sentiment of what you're saying, if this should be good to go, any excuse to be reminded of it is good enough.

We've been wanting to do *something* about colour handling in OBS for over a year, and this is a great start. That said, do you feel that you have bigger priorities you're tackling currently? If so, I suppose it can wait a few more months.

jpark37:
On OBS I don't, but even if this is approved, I may not circle back to the PR right away because I have another project that I'm trying to get off the ground.

I was more commenting to Jim that he probably shouldn't drop whatever he's working on to review the RFC again.

jpark37:
This is only tangentially relevant since I wasn't planning to implement color conversions and only fix metadata tags, but after speaking with @haasn from mpv, and thinking about it some more, I think aliasing 709 as sRGB is more correct than I thought.

- Nearly everyone does this, so that alone kind of makes it true.
- If we think of 709 data as being destined for 1886 displays, and sRGB data as targeted for sRGB displays. Aliasing 709 as sRGB essentially darkens the mid-tones, but displaying them on an sRGB display instead of 1886 brightens them. In other words, 709 to sRGB color gamma (~1.956 to ~2.2) kind of cancels 1886 to sRGB display gamma (2.4 to ~2.2). They aren't strictly equivalent, but probably close enough.

ifohancroft:
I can test the different color spaces and things, under Linux, if someone was to provide test files and maybe a short description of how to tesf/check results.

haasn:
This issue makes numerous references to sYCC as being "the equivalent of BT.709 but with sRGB transfer function". This does not appear to be the case according to IEC61966-2-1's definition of "sYCC" which uses **BT.601 matrix coefficients**.

So to be clear, when I said that it makes sense to have OBS default to BT.709+sRGB, I specifically meant BT.709+sRGB, not sYCC (i.e. BT.601+sRGB).

I recommend keeping this proposal as is but changing the term 'sYCC' to something that clearly distinguishes it from the other usage of sYCC in the literature, even if that just ends up being 'sYCC-709' or something.

jpark37:
Oops, I missed that somehow. I'll update everything. I agree that defaulting to 709 matrix coefficients is still preferable.

jpark37:
Okay, went through and scrubbed all incorrect usages of sYCC. Everything should be in order now.

jpark37:
@haasn Wait a second... I'm looking at the H.264 spec, and it says sYCC is KR = 0.2126; KB = 0.0722. (I'm not as stupid as I thought!) The H.265 spec does have KR = 0.299; KB = 0.114 though.

I'm not sure if the discrepancy is because one of the other is wrong because I don't want to pay for IEC 61966-2-1. Either way, I'm going to leave the sYCC term out of OBS.

Edit: More digging seems to indicate the H.264 spec is wrong, but can you really blame me for believing them?

kodawah:
For the record, I believe that h264 is going to be fixed in the next release (see AJ0023 if you have access).

> Edit: More digging seems to indicate the H.264 spec is wrong, but can you really blame me for believing them?

¯\\\_(ツ)\_/¯ this happens when people put specifications behind a paywall

haasn:
A good enough reason to clearly disambiguate our terms regardless :)

jp9000:
Approved -- if anyone has any objections to this and thinks it should be reverted, now is your time to speak up. I'm still new to the RFC process so I don't think I was supposed to explicitly merge right away, but I didn't realize that until after the fact.

So instead for this RFC if anyone has any objections to this RFC now is the time to speak up.

Fenrirthviti:
The merge can be reverted so not a big deal, final comment period has started.

Fenrirthviti:
FCP has finished and this can be considered complete.

jpark37:
It looks like YouTube changed their video pipeline from when I tested in January to actually apply gamma transformation to their video. What this means is that they degamma with sRGB and gamma with 709, effectively shrinking the midtones. Since most video players will alias 709 data to sRGB, the result is too dark.

Given this, I think 709 is probably the better default.

RytoEX:
> Either way, I'm going to leave the sYCC term out of OBS.

It's a bit late to notice this, but the commit message for the RFC still says sYCC instead of sRGB.

> Add RFC: sYCC Color Space setting

jpark37:
I blame Linus. A real source control system would let people fix their commit messages.

RytoEX:
Technically we could, but at the risk of breaking any existing checkouts.  I have no strong feelings either way, but it's probably not terribly important to fix the commit message.

haasn:
`git notes`?

jpark37:
I was mostly joking guys.

@haasn Google/YouTube are deciding the fate of the universe, well, 709/sRGB handling, and I thought you might be interested: https://bugs.chromium.org/p/chromium/issues/detail?id=947798#c25

kasper93:
First of all thanks for switching to BT.709 by default, it makes things a lot better. But I think the gamma issue is up in the air...

> If we think of 709 data as being destined for 1886 displays, and sRGB data as targeted for sRGB displays. Aliasing 709 as sRGB essentially darkens the mid-tones, but displaying them on an sRGB display instead of 1886 brightens them. In other words, 709 to sRGB color gamma (~1.956 to ~2.2) kind of cancels 1886 to sRGB display gamma (2.4 to ~2.2). They aren't strictly equivalent, but probably close enough.

@jpark37: So by lying in file metadata you exploit display gamma response to convert roughly back to "proper" brightness. Now what if someone have BT.1886 calibrated display for movie watching? Let say someone want to watch Twitch on TV? Or has proper video pipeline that in fact converts color space and gamma to target display? Assumption that everyone is sRGB is not that correct.

I know that current state of industry is not perfect. But basing encoding decision on lacking playback options makes things only worse. Because now that all content is mislabeled it is impossible to correctly support down the line ever. Contrary if we push proper video streams it is on the processing/playback side to adjust to correctly support it.

Also on the pragmatic side, most people doesn't care/notice color mismatch unless you show them A/B comparison or something is extremely wrong. Let's say the transition period where videos would be little bit darker would not be noticeable for most (well people watch movies meant for bt.1886 displays on sRGB everyday). Most have shitty displays where colors, white point and gamma are off from sane values anyway... And for those who have calibrated displays proper video files are key to view them correctly.

jpark37:
We're not making the decision though. It's a user setting. We just picked a reasonable default. They can pick whatever they deem correct.

ifohancroft:
Here's my take on this:

Allow the user (possibly under Advanced Settings) to choose whether OBS should:
 - Do nothing about the color space (whether that's conversion or tagging) and just pass it through as whatever it comes as or claims to come as
 - Mark/tag the signal as being particular color space (with color spaces to choose from)
 - Actually convert the color space (with color spaces to choose from)
 - Perhaps a fallback/default would be the current way it works or the way this spec suggests it should work

If you don't know or care about color spaces, you wouldn't touch the option, but if you do, you'd want to have control over what happens (specially if you have gone through the trouble of learning about color spaces, choosing one to use, matched your cameras and equipment and made sure you are using the color space and workflow for it all the way through) and I feel like my suggestion of where you can choose whether OBS does nothing, just marks the signal as something or actually converts it is gives maximum control.

Even without the conversion, my main point with this comment is: There needs to be an option to tell OBS to not care about the color space and just pass it as what it is and an option to tell it to pass it as a specific color space, and maybe the fallback/backwards compatible of working the way it currently does/the spec suggests it should.

kasper93:
> We're not making the decision though. It's a user setting. We just picked a reasonable default. They can pick whatever they deem correct.

@jpark37: Of course you do. Most of OBS users have no idea what this option do and what are implications. And even with documentation explaining this option, always recommended default will be used by most. Which happens to be incorrect.

It is also not clear that this option does set only metadata and **do not** do proper conversion. I mean if it is user setting, it should convert source to proper colorspace that match metadata, so the resulting video stream is correctly tagged.

> Here's my take on this:

@ifohancroft: Agree, but I wouldn't make it too complicated. I think we need to remember that most users doesn't understand this stuff and it is on software to prevent misconfiguration by making configuration easy and without unsaid assumptions.

I don't know how are sources handled and composited, but let's look at it from end of the pipeline. 
* Output metadata should match output stream, so if needed the conversion should be done.
* Output metadata could have as-is option, meaning the tags would follow whatever is on input, without conversion.
* Each source should have its options so it is know how it should be interpreted.

But like I said I don't know how OBS composite different sources and so on. The current most immediate issue I see is that output colorspace settings doesn't do conversion and default values are not matching most common usecase of capturing pc screen.

I didn't want to start some unneeded discussion, but was configuring my video playback pipeline and right now I need to switch profiles just for Twitch manually, because of this slight metadata mixup and I blame OBS for that :) But I guess one can't have everything.

jpark37:
Can you explain a bit more about why you have a separate profile for Twitch? What are your differences?

jpark37:
There's a couple reasons I don't want to get into the business of color space data conversion.

- It's complicated for users, and they mostly don't care. As we have tested, video players mostly don't care. I don't think it's a good idea to make the functionality theoretically correct (and this is still up for debate) if it's just going to play back with unintended colors. If a user really cares, they can use the LUT filter, or use/write a color-conversion filter plugin.
- sRGB to 709 data conversion will crush blacks at about a 3:1 ratio (12.92 / 4.5). Do we really want to add code that just replaces one wrong (incorrect absolute value) with a different wrong (banding, detail loss)?

I've heard that YouTube wants Chromium to make 709/sRGB conversions happen across platforms. If this takes place, there's more incentive to offer a form of streamlined, optional conversion functionality, but I'd rather not go through that design exercise until it's necessary.

kasper93:
Sorry, for delayed response, I got tangled in other things :)

> Can you explain a bit more about why you have a separate profile for Twitch? What are your differences?

To make it short. I have 3DLUT generated for my display which I use to target BT.1886 gamma for BT.709 content. Now because twitch streams are mislabeled as BT.709 instead having sRGB transfer function. I cannot automatically do proper conversion. So in the end I need to manually change the settings. 

> It's complicated for users, and they mostly don't care. As we have tested, video players mostly don't care. I don't think it's a good idea to make the functionality theoretically correct (and this is still up for debate) if it's just going to play back with unintended colors.

The point of making it theoretically correct is that when someone wants to care there will be proper metadata available. Of course if one don't care it doesn't really matter. But we need to start from the source making things correct. Else everything down the line will be forced to "not care" or "ignore metadata", because metadata is incorrect. And by making things correct you shift focus to actual players that possibly playback it incorrectly. I know compatibility is a tough issue, but there is metadata and standards to talk with each other effectively. 

> There's a couple reasons I don't want to get into the business of color space data conversion.

I kind of agree, color space conversions could be tricky. My main point is that produced files should have metadata that properly reflects the actual data. And to achieve that there will be probably needed some conversions of the capture input that are encoded. Of course it doesn't have to be 100% perfect, but let say the main target which is PC screen capture should probably use sRGB.

jpark37:
I came across this post recently, and apparently you're not supposed to mess with the color values when converting between sRGB and Rec. 709, which would mean what YouTube is doing is wrong.

https://community.acescentral.com/t/relationship-between-srgb-bt-1886-and-rec-709/4036/2
> sRGB was created to give the same appearance as Rec.709 for same code values, but in a different viewing condition (lit office instead of darker living room). Charles Poynton can give you the full story AFAIK.
That is, a Rec.709 video played on an sRGB display (unconverted) in the office “will appear similar” to same video played in the living room on a BT.1886 display.
This means that there shall be no value conversion between Rec.709 values and sRGB values.

It seems like this was a design goal of sRGB the start, so OBS's "do-nothing" approach to the color values was the right thing to do all along.

PR #11, opened 2020-01-24
Add RFC: Downstream keyer

cg2121:
### Description
Add the ability for users to create global scenes thats are on top of every scene.

### Motivation and Context
Users won't have to create meta scenes and copy them to every scene.

### Link
https://github.com/cg2121/rfcs/blob/downstream-keyer/accepted/0011-downstream-keyer.md

### DIFF
--- a/accepted/0011-downstream-keyer.md
+++ b/accepted/0011-downstream-keyer.md
@@ -1,0 +1,39 @@
+- Start Date: 2020-01-23
+- RFC PR: #11
+- Mantis Issue: https://obsproject.com/mantis/view.php?id=602
+
+- Related PRs:
+	1. https://github.com/obsproject/obs-studio/pull/1321
+
+# Summary
+
+Add the ability for users to create global scenes thats are on top of every scene.
+
+# Motivation
+
+Users won't have to create meta scenes and copy them to every scene.
+
+# Detailed design
+
+## User UX
+
+- Have an extra dock for the DSK scene (same as the scenes dock)
+- Have the dock on by default, so users easily know about it
+- Users could switch between DSK channels with combo box in the dock
+- Alternatively, use hotkeys to change DSK channels, so we don't clog up the UI
+- There would be an extra option in the combo box for disabling the DSK
+- Call the DSK "Global Scenes" so users would more likely know what it is
+
+## Other
+
+- PR only implemented single DSK channel
+- In the PR, users could disable the DSK with hotkeys
+- PR needs to be refactored to remove libobs usage
+
+# Drawbacks
+
+Users might not know what a DSK is.
+
+# Alternatives
+
+Add an existing scene to scenes.


ryantheleach:
Personally I fail to understand what advantage global scenes gives over meta-scenes, other then taking up more real estate. Maybe it's because I don't understand the channels concept, or how the UX improves (It just looks like crowding to me)

In addition the terminology `DSK` used in the RFC may be familiar to those from a broadcast background, But I've only encountered the term once or twice in my research into doing this as a hobby, and even then it seems/feels dated, much the same way as applications that still use a floppy disk icon for save.

Global Scenes makes much more sense, But I again have to ask, what UX is achieved by adding it?

If it's a matter of viewing 'sources' for 'non current' scenes-as-sources / shared meta scene, I'd argue that maybe it should be possible to create 'source' dock panels for all scene-as-sources instead, and let the user work out how they would like it handled, and on which scenes they should be visible.

Global Scene's as described in the RFC seems like a compromise that gives a 'simple' solution, to a 'complex' request, that will only serve to confuse newer users, and frustrate power users, unless I'm missing something fundamental.

I'd be interested in seeing some mockups, to make sure that my understanding from the text is sufficient.

As an aside, I much prefer the 'pinning' of sources, for a **simplified** global source (mentioned in the PR), instead of using meta-scenes. But it's clearly not going to cover all use-cases as it's on **all** scenes. However Pinned Sources should have an additional Visibility button, that toggles the visibility, not globally, but per scene.

kkartaltepe:
RFC is missing a how to teach section which might help with the above concerns.

RFC proposes multiple channels but does not mention any method of synchronizing channel to scene changes. I think thats likely to be a big issue if we support multiple channels (or a non-issue if we stick to single channel). Meta sources solve that issue at least, and clever hotkeying can let you change the active metasource while live.

accepted/0011-downstream-keyer.md
```diff
@@ -0,0 +1,39 @@
...
+
+# Motivation
+
+Users won't have to create meta scenes and copy them to every scene.
+
+# Detailed design
+
+## User UX
+
+- Have an extra dock for the DSK scene (same as the scenes dock)
```
jp9000, line 20: I agree with this, and it should be treated like a secondary scene list, probably with its own transitioning.  Question is how to handling transitioning, especially considering people might want to use different transitions for DSKs.

accepted/0011-downstream-keyer.md
```diff
@@ -0,0 +1,39 @@
...
+- Have the dock on by default, so users easily know about it
```
jp9000, line 21: I don't agree with this, I think this goes against the current design pattern where advanced features are hidden/separated by default.

accepted/0011-downstream-keyer.md
```diff
@@ -0,0 +1,39 @@
...
+- Users could switch between DSK channels with combo box in the dock
```
jp9000, line 22: Personally, I would prefer to just have a DSK 1 scene list, DSK 2 scene list or something like that.  Probably even separate docks for them.  I'm not sure how others would feel about that idea however.  @dodgepong might have some strong opinions here.

accepted/0011-downstream-keyer.md
```diff
@@ -0,0 +1,39 @@
...
+- Alternatively, use hotkeys to change DSK channels, so we don't clog up the UI
+- There would be an extra option in the combo box for disabling the DSK
```
jp9000, line 24: Yea, having a "None" option for the DSK which it would be set to by default or something would likely be necessary.

accepted/0011-downstream-keyer.md
```diff
@@ -0,0 +1,39 @@
...
+- Call the DSK "Global Scenes" so users would more likely know what it is
```
jp9000, line 25: Because it's more of an advanced feature, I don't think we necessarily need to dumb down the name.

accepted/0011-downstream-keyer.md
unknowndomain: I think the issue is that the DSK is functionally a clone of the Scenes list so you can only show one at a time, where you'd reasonably expect to be able to toggle DSK 1, 2, 3 on and off like muting or unmuting audio sources.

accepted/0011-downstream-keyer.md
unknowndomain: I agree it's a more advanced feature but when users realise this is the way to do channel bugs etc. I suspect it will get more use, and thats more likely if it's called global rather than DSK.

It might be interesting in future to add a notification when a user adds a source to multiple scenes (3+) the first time to show a dialog that tells them about global scenes and offers to enable it for them.

accepted/0011-downstream-keyer.md
```diff
@@ -0,0 +1,39 @@
...
+- Related PRs:
+	1. https://github.com/obsproject/obs-studio/pull/1321
+
+# Summary
+
+Add the ability for users to create global scenes thats are on top of every scene.
+
+# Motivation
+
+Users won't have to create meta scenes and copy them to every scene.
```
summershrimp, line 14: My opinion:
```
# Motivation

1. Users won't have to create meta scenes and copy them to every scene.
2. The global scene won't affected by scenes transition effect.
```

accepted/0011-downstream-keyer.md
summershrimp: I think it should be called "(Active) Output Overlay Scene" or "(Active) Global Overlay Scene".

accepted/0011-downstream-keyer.md
summershrimp: I think DSK should reuse the original scenes, may have separate transition effect. So we needn't make a whole new scene lists for them, we can just add a "Transit to Global Overlay" option to scenes' right-click menu. And add a "Global Overlay Transition Effect" selector below the original scenes' transition effect selector, or may just use current active transition effect. In this way, it's more like a traditional hardware switcher.
Using a new Scenes list for DSK is also acceptable.

summershrimp (commented):
> I think DSK should reuse the original scenes, may have separate transition effect. So we needn't to make a whole new scene lists for them, we can just add a "Transit to Global Overlay" option to scenes' right-click menu. And add a "Global Overlay Transition Effect" selector below the original scenes' transition effect selector, or may just use current active transition effect. In this way, it's more like a traditional hardware switcher.


In this way, we won't change the UI too much, and it will be ignored for those who doesn't need this feature.

DSK is a really necessary function for live production makers, please consider raise the priority.

Fenrirthviti:
After a lot of internal discussion, I think this approach is slightly confusing and a little more complicated than it needs to be. For an MVP for DSK, or really, DSK-like functionality as replicating DSKs 1:1 is a bad idea for us, here is what we think would be best as a workflow:

- Create a scene, add sources to that scene to your hearts content
- "Pin" the scene as the top layer (either through an icon next to the source or the context menu)
- For first iteration, only one scene is allowed to be "pinned"

OPTIONAL NICE TO HAVES:
- Allow a toggle for on top of or below normal transitions
- Allow a transition if switching which DSK is "pinned"
- Allow for editing a scene while it is already layered on top
- Hotkeys

An example UI that shows (potentially) multiple pinned scenes:
benexample

And an example that shows how one might potentially pin a scene using the context menu:
https://cdn.discordapp.com/attachments/352910054208307211/802062309471158282/CrappyMockup.mp4

I apologize for posting this feedback so late.

PR #13, opened 2020-03-05
Add RFC: Interlaced output

mufunyo:
### Description
Add the ability for OBS to output interlaced video.

### Motivation and Context
TV broadcast is often interlaced, and Decklink cards offer the ability to output interlaced signals, but due to a limitation in OBS' output design, Decklink interlaced output is reduced to half frame rate progressive in an interlaced signal (i.e. 30p in 60i).

### Link
https://github.com/mufunyo/obs-rfcs/blob/interlaced-output/accepted/0013-interlaced-output.md

### DIFF
--- a/accepted/0013-interlaced-output.md
+++ b/accepted/0013-interlaced-output.md
@@ -1,0 +1,37 @@
+- Start Date: 2020-03-05
+- RFC PR: #13
+- Related Mantis Issue: https://obsproject.com/mantis/view.php?id=1516
+
+# Summary
+
+Add the ability to output interlaced video.
+
+# Motivation
+
+TV broadcast is often interlaced, and Decklink cards offer the ability to output interlaced signals, but due to a limitation in OBS' output design, Decklink interlaced output is reduced to half frame rate progressive in an interlaced signal (i.e. 30p in 60i).
+
+# Detailed design
+
+## Technical considerations
+
+- The output pipeline should be so that interlacing is applied after output scaling
+- Output YUV shaders should account for interlaced chroma sample positions
+
+## User UX
+
+- Allow choosing top field first / bottom field first
+- Add a global toggle to enable the option to select interlaced output in advanced options, to prevent unintentional use
+
+## Other
+
+- Potentially allow using a vertical lowpass filter to prevent flicker
+- Devise a way to ensure deinterlaced sources preserve their field order to prevent outputting interpolated lines instead of original source lines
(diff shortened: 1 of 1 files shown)


jp9000:
Hm, so is this specifically for the decklink output, or for recording as well?

mufunyo:
Decklink, recording, and streaming, yes. Should also be good for SRT streaming.

GrandyB:
+1 for this - working in esports and in talks with TV companies, 1080i50 is the format they're looking for. I think vMix may have this functionality, but it'd be great if OBS did.

mufunyo:
> +1 for this - working in esports and in talks with TV companies, 1080i50 is the format they're looking for. I think vMix may have this functionality, but it'd be great if OBS did.

It may be too long for you to wait for this RFC to be approved and fulfilled, so I would recommend using OBS Decklink output to produce a compliant 1080p50 output, and then daisy chaining in a scan converter (Blackmagic has affordable models) to bring it down to 1080i50 for broadcast. If you want to discuss further options we shouldn't clutter this RFC - you can find me in the OBS Discord.

tt2468:
I know it's been a few years since the last activity on this thread. I hate to be a downer, but I'm not sure what reasonable justification there would be behind adding an interlaced mode at this point. Interlaced content is arguably going away gradually. To me, it would almost be cruel to implement this so late in its lifetime. Scan converters are also definitely much more accessible. I think the MD-HX supports p->i conversion.

Fenrirthviti:
While we understand that this might be useful for some edge cases or specific communities, we also feel that there are other tools and solutions that accomplish this, and the added support and maintenance cost for this is not worth it for us.

Moving to Final Comment Period, with Disposition Close

harrypm:
Interlacing support is a core feature of FFmpeg. 

Considering how many pretty much clueless idiots use OBS to digitise SD/HDV interlaced footage and are crunching it into progressive with limited filters offered when they don't need to.

Alongside the clear and obvious support of something called "broadcaster" not supporting a basic broadcast feature this should be reopened and implemented because it's a zero maintenance addition to allow for direct interlaced encoding.

Many users have hacked around this problem many times with just updated FFmpeg builds literally that's the only change that needs to be made the single GUI element forcing a limited parameter set for ensuring correct interlaced flagging that's it.

- The FM RF Archival Community

Fenrirthviti:
OBS is not a frontend for FFmpeg, so what FFmpeg does or does not support as core features is largely irrelevant to OBS.

PR #18, opened 2020-04-18
Add RFC: Notification widget

cg2121:
### Description
Add the ability to display notifications to the user.

Screenshot from 2020-04-17 20-49-50

### Motivation and Context
Currenty, there is no way of showing custom notifications to users.

### Link to RFC
https://github.com/cg2121/rfcs/blob/notify-widget/accepted/0018-notification-widget.md

### DIFF
--- a/accepted/0018-notification-widget.md
+++ b/accepted/0018-notification-widget.md
@@ -1,0 +1,35 @@
+- Start Date: 2020-04-17
+- RFC PR: #18
+- Mantis Issue: N/A
+
+- Related PRs:
+	1. https://github.com/obsproject/obs-studio/pull/1832
+
+# Summary
+
+Add the ability to display notifications to the user.
+
+# Motivation
+
+Currenty, there is no way of showing custom notifications to users.
+
+# Detailed design
+
+## User UX
+
+- Display notifications at top of main window
+- Have three notification types (Info, Error, and Warning)
+- Each type will display a different icon and the widget would be a different color for each
+- Display custom text
+- Add button for different types of actions (open source properties, open settings window, open url)
+- The button's text will be customizable
+- The widget will be floating over the window so other widgets aren't resized when it is displayed
+- Add api to libobs so plugins can display notifications, instead of just the frontend
+
+# Drawbacks
+
+Users might find the notifications annoying.
+
+# Alternatives
+
+Display warnings in status bar.


accepted/0018-notification-widget.md
```diff
@@ -0,0 +1,35 @@
...
+
+## User UX
+
+- Display notifications at top of main window
+- Have three notification types (Info, Error, and Warning)
+- Each type will display a different icon and the widget would be a different color for each
+- Display custom text
+- Add button for different types of actions (open source properties, open settings window, open url)
+- The button's text will be customizable
+- The widget will be floating over the window so other widgets aren't resized when it is displayed
```
dodgepong, line 26: I think this is a double-edged sword, here. It's nice that it doesn't resize the preview, but it's not nice that it obscures the preview. I'm still of a mind that notifications should be placed in the status bar somehow, with an option to click on the notifications to expand them or pop up a larger window.

Fenrirthviti: I agree from a informational notification standpoint, but anything important enough to be sent as a warning or error SHOULD disrupt the user and get in the way so that they can't miss it. This isn't something we should be hiding behind an optional click, as I feel that defeats the point.

energizerfellow: I agree that the notifications should be out, loud, proud, and distracting by design to get user's attention.  Many of OBS's (admittedly limited) UI notifications are, if anything, too subtle for a large chunk of the user base.

That said, however, I do like the idea of having a (non-default) UI option to consolidate the notifications into something like Windows 10's notifications where there's a number and the user can click to get a list of notifications. Bonus points if they're color coded in the list that comes up.

palana:
BTW what's the difference in goal between this RFC and https://github.com/obsproject/rfcs/pull/19? Is this supposed to be about just the UI portion, where the other RFC has more details on actual internals?

VodBox:
> BTW what's the difference in goal between this RFC and #19? Is this supposed to be about just the UI portion, where the other RFC has more details on actual internals?

@palana They are currently at the moment competing RFCs. As neither have been accepted yet, we have the opportunity to discuss them both and compare them.

accepted/0018-notification-widget.md
```diff
@@ -0,0 +1,35 @@
...
+- Add api to libobs so plugins can display notifications, instead of just the frontend
```
palana, line 27: Not sure I agree that this (libobs API) is a UX issue necessarily, but could you add some detail on what this should look like and how this might influence your prototype?

accepted/0018-notification-widget.md
```diff
@@ -0,0 +1,35 @@
...
+# Motivation
+
+Currenty, there is no way of showing custom notifications to users.
+
+# Detailed design
+
+## User UX
+
+- Display notifications at top of main window
+- Have three notification types (Info, Error, and Warning)
```
palana, line 21: Could you add some details to this RFC on how your PR https://github.com/obsproject/obs-studio/pull/1832 used these, and what you'd think these could be used for? E.g. when would you use an "Info" notification (concrete example please, not just "something that happens but doesn't require immediate user action")

accepted/0018-notification-widget.md
```diff
@@ -0,0 +1,35 @@
...
+- Each type will display a different icon and the widget would be a different color for each
+- Display custom text
+- Add button for different types of actions (open source properties, open settings window, open url)
+- The button's text will be customizable
+- The widget will be floating over the window so other widgets aren't resized when it is displayed
+- Add api to libobs so plugins can display notifications, instead of just the frontend
+
+# Drawbacks
+
+Users might find the notifications annoying.
```
palana, line 31: Anything that can be done to mitigate this?

accepted/0018-notification-widget.md
```diff
@@ -0,0 +1,35 @@
...
+
+# Alternatives
+
+Display warnings in status bar.
```
palana, line 35: Can you lay out some reasoning as to why you think the notification widget is desirable/should be preferred over the status bar?

accepted/0018-notification-widget.md
```diff
@@ -0,0 +1,35 @@
...
+
+# Detailed design
+
+## User UX
+
+- Display notifications at top of main window
+- Have three notification types (Info, Error, and Warning)
+- Each type will display a different icon and the widget would be a different color for each
+- Display custom text
+- Add button for different types of actions (open source properties, open settings window, open url)
```
palana, line 24: Your PR only seems to use this to open a URL instead of taking internal actions; any thoughts on how to both present a URL "button" ("click here to learn more about this issues") and a "fix this" button? Are there use-cases that don't need a button?

Also, should there be a limit on amount of text displayed (does the widget support multiple lines)?

palana (commented):
Not sure if this RFC is intended to also discuss expiring notifications and a few other things that https://github.com/obsproject/rfcs/pull/19 discusses; as is, I would encourage collaboration with https://github.com/obsproject/rfcs/pull/19, so we'd potentially get a more complete RFC with discussion on different approaches and trade-offs

Also, I'd suggest "embedding" screenshot in the actual RFC

accepted/0018-notification-widget.md
unknowndomain: Would use the operating systems existing notification systems not be a good place to start with this, both Windows and macOS already have good systems which users are familiar with and would avoid reinventing the wheel over how to design this.

cg2121:
Closing because #19 is basically the same as this and has a lot better details. And having this one open would be a waste of time.

PR #19, opened 2020-04-18
Add RFC: App Notification/Alert System

WizardCM:
### Description

A method for plugins and scripts to notify the user about important or time-critical information. I'll use the term "Alerts" for the rest of the RFC to cover this functionality. This RFC is mostly around the backend design, with elements of the frontend and UX considerations.

The end goal is an alert widget of some description, with a panel/window for in-depth and historical alerts.

### Motivation and Context

Internally, OBS does a lot of things at once, and any one of them can fail for a variety of reasons. Currently, the only way for a plugin to document that something went wrong is to log it. Then, someone with more experience requests and reads the log, to then provide the user with an actual solution. This is a whole lot of unnecessary work for both the user (who is more likely to give up) and our support team.

### [View the RFC](https://github.com/WizardCM/obs-rfcs/blob/app-notifications/accepted/0019-app-notifications.md#summary)

### DIFF
--- a/accepted/0019-app-notifications.md
+++ b/accepted/0019-app-notifications.md
@@ -1,0 +1,111 @@
+- Start Date: 2020-04-17
+- RFC PR: #19
+- Mantis Issue: N/A
+
+# Summary
+
+A method for plugins and scripts to notify the user about important or time-critical information. I'll use the term "Alerts" for the rest of the RFC to cover this functionality. This RFC is mostly around the backend design, with elements of the frontend and UX considerations.
+
+The end goal is an alert widget of some description, with a panel/window for in-depth and historical alerts.
+
+# Motivation
+
+Internally, OBS does a lot of things at once, and any one of them can fail for a variety of reasons. Currently, the only way for a plugin to document that something went wrong is to log it. Then, someone with more experience requests and reads the log, to then provide the user with an actual solution. This is a whole lot of unnecessary work for both the user (who is more likely to give up) and our support team.
+
+## Example Alerts
+
+> **Some sources failed to load.** It looks like a plugin may be missing. <a href="#">Read more</a>.
+
+> **Some sources are pointing to missing files.** They may have been moved or deleted. <a href="#">List problematic sources</a>.
+
+> **Output aspect ratios don't match.** This can result in stretched recordings. <a href="#">Fix it</a>.
+
+> **Mismatched Sample Rate found** This can result in audio desync, crackling, etc. <a href="#">Fix it</a>.
+
+> **You just finished a 4 hour stream.** We detected some issues that can be fixed. <a href="#">Check via the analyzer</a>.
+
+> **You were disconnected from your streaming service.** You can read our <a href="#">troubleshooting guide</a> for common solutions.
+
+> **OBS is having trouble maintaining 60 fps.** You can solve this by running OBS as Administrator. <a href="#">Not sure how?</a>
+
+> **An OBS update is available!** You can update manually at any time. Read the full <a href="#">list of changes here</a>.
+
+> **Display Capture requires extra configuration.** You can follow our <a href="#">Laptop Guide here</a>.
+
+> **Issues going live?** You can keep an eye on <a href="#">@TwitchSupport</a> for realtime updates.
+
+# Detailed design
+
+Alerts can have multiple levels based on urgency. Critical, Warning, and Info. In the main window, Alerts are shown one at a time rather than stacking. They can either be dismissed automatically (for example, when an issue goes away or the user fixes the issue manually), or by dismissing it.
+
+## UX Notes
+
+Visually, the Alert will have a coloured background depending on the priority, and within the container an icon to represent the priority, the summary, an 'expand' button to show the full description, and a dismiss button. If expanded, we can provide a direct link to a full list of Alerts. This could either be shown above the preview, or even as part of the status bar.
+
+A full, scrollable, list of active Alerts can be viewed in a secondary window, including Alerts that the user dismissed but are still active (as in, the problem hasn't been solved). From here, the user can explicitly decide to enable/disable alerts from specific plugins. For example, "never show invalid audio device errors". This page could also have methods of filtering/sorting by priority/date, etc.
+
+The summary should be capable of HTML-anchor style actions, allowing the capability to open external webpages or even open specific windows within OBS - say, the Properties window for a specific source, or the Settings window pointed to a specific tab with a certain setting highlighted.
+
+Note: This Alert system would not expose a way to pop up warning/error dialogs. That is not its purpose.
+
+## Internals
+
+The Alerts framework would likely have to be built as part of libobs, providing a pathway from plugins to the UI.
+
+When generating an Alert, the following (bare minimum) parameters should be available:
+
+```js
+{
+  "timestamp": 1585817608, // An epoch timestamp of when the issue occurred. Generated automatically
+  "priority": "CRITICAL", // Enum of CRITICAL, WARNING, INFO // Required? Could default to warning?
+  "summary": "", // Required. HTML-anchor capable. A short summary of the issue, visible from the main window
+  "description": "", // Optional. A lengthy, more informative HTML string providing more context and inline solutions. Might support newlines and dot points
+  "persist": true, // Optional, default to false. If the user dismisses the alert, whether it should persist in the full list. Useful for alerts that are tied to a fixable issue.
+}
+```
+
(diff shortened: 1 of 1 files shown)


accepted/0019-app-notifications.md
```diff
@@ -0,0 +1,70 @@
...
+
+# Additional Information
+
+There are many things to consider in this design. It's highly likely that the backend & UI would be developed & PR'd independently rather than together. This'd ensure a solid backend design, and an easier time reviewing it.
+
+The backend design & scope should be fleshed out significantly more before a real impelementation is started.
+
+# Unresolved questions
+
+* Should Alerts have a timeout? A list of methods to dismiss (scene collection switch, profile switch)?
```
palana: We can probably "solve" that second part to some degree by following the split that e.g. the hotkey system follows: `obs_add_source_notification`/`obs_add_frontend_notification`/`obs_add_output_notification`

Neither scene collection nor profile are libobs concepts, they're UI concepts, so exposing those in libobs is not a good idea imo. Instead what libobs can do is dismiss notifications from sources/outputs/services/… that get removed

So for example an image source notification "could not open file xyz.jpg" would go away automatically once that source is removed

I can't think of examples of where I'd want notifications to have a timeout, and the more I think about it the more I'd want all notifications to be logged (to prevent the whole "phone made a notification sound while nothing whatsoever was going on, but when checking the phone it didn't display anything, and there was no record of what had happened"), but this is likely going to cause fatigue issues very fast if there's no "sane" bulk dismiss option

palana (commented):
Can you give a few examples (in the RFC) of what kind of notifications you'd want to enable with this? E.g. things that seem interesting to me (but maybe not to other people):

- Sources/outputs/filters not being loaded (dlls etc missing?)
- Sources missing their referenced files
- Recording running out of storage space(?) (e.g. should this be dismissed when the recording is stopped? Restarting the recording immediately is likely going to run into the same issue, and changing just the file name (not the path) will potentially also run into the same issue, even when switching between "simple" and "advanced" outputs)
- Aspect ratio mismatch between canvas and output resolution (while not a huge concern right now necessarily, people can already edit the relevant profile files and run into issues; adding separate output profiles or similar might increase the likelihood of this happening)

RytoEX:
Is there a reason we cannot use showMessage from QSystemTrayIcon?  We already do this in a few places.  We could probably expose a function for plugins to just pass messages to it.

WizardCM:
Per the above feedback, I've updated the RFC with the following:

* Examples of potential official uses of the Alerts system (plugins could do other things)
* Included @Palana's example for a potential API, and documented why a timeout is a bad idea
* Documented the alternative suggested by @RytoEX, and its upsides/downsides

Keep the feedback & questions coming. :)

accepted/0019-app-notifications.md
```diff
@@ -0,0 +1,111 @@
...
+
+There are many things to consider in this design. It's highly likely that the backend & UI would be developed & PR'd independently rather than together. This'd ensure a solid backend design, and an easier time reviewing it.
+
+The backend design & scope should be fleshed out significantly more before a real impelementation is started.
+
+# Alternatives
+
+- Exposing QSystemTrayIcon's `showMessage` - providing plugins a way to use native notifications.
+  - Downsides include:
+    - They will trigger a sound that the user cannot mute
```
RytoEX, line 97: System Sounds can be muted by:
* muting System Sounds in the Volume Mixer
* setting the Notification sound to None
  * Settings > System > Sound > Related Settings > Sound Control Panel > Sounds tab)
  * Settings > Personalization > Themes > Sounds)
* Enabling Focus Assist with "Priority only" or "Alarms only"

You can also disable or customize notifications sent from OBS in Settings > System > Notifications & actions > Get notifications from these senders.

accepted/0019-app-notifications.md
```diff
@@ -0,0 +1,111 @@
...
+    - Windows automatically disables these when a game is running (via Focus assist)
```
RytoEX, line 98: The "When I'm playing a game" is a Focus Assist rule that can be disabled.  Also, you can add OBS to the Focus Assist Priority List to exclude it from this rule/feature.

accepted/0019-app-notifications.md
```diff
@@ -0,0 +1,111 @@
...
+    - they are distracting and interrupt the user
```
RytoEX, line 99: See previous comments about how these notifications can be customized via Windows Settings.

accepted/0019-app-notifications.md
```diff
@@ -0,0 +1,111 @@
...
+    - Due to how they're visually exposed to the user, they could be abused by plugins ("annoying")
```
RytoEX, line 100: This same downside can apply to any notification method we create ourselves, unless we also build in a way to detect and suppress notifications that we classify as "spam" or "too frequent".

RytoEX (commented):
I did some investigating on the downsides listed for using QSystemTrayIcon's `showMessage`, at least on Windows 10.  I don't know what notification customization is available on other OS's.

As it turns out, we already do have a public function for exposing this:  OBSBasic::SysTrayNotify.  OBSBasic::SysTrayNotify is declared in [window-basic-main.hpp](https://github.com/obsproject/obs-studio/blob/e822b474276838f959ecb2f15cffd18b7e721863/UI/window-basic-main.hpp#L720) and defined in [window-basic-main.cpp](https://github.com/obsproject/obs-studio/blob/4ed3df14a624dbcc41abbe4e762aade184762891/UI/window-basic-main.cpp#L7216).  It is also used throughout our codebase already.

RytoEX:
It looks like GitHub ate the image I posted as a reference for customizing notifications on Windows 10.

You can also disable or customize notifications sent from OBS in Settings > System > Notifications & actions > Get notifications from these senders:
image

accepted/0019-app-notifications.md
WizardCM: By disabling that focus assist rule, the user has to open up their system to other notifications, including chat applications, standard windows alerts (like Windows Update), and more. I do not see this as a solution.

accepted/0019-app-notifications.md
WizardCM: The downside of this is that the user is then disabling standard OBS notifications - like the current "Disconnected. Reconnecting.." one because they want to quiet other notifications.

On top of that, they don't visually persist, which is the upside of the proposed system.

accepted/0019-app-notifications.md
RytoEX: You can add OBS to the Focus Assist Priority List.

accepted/0019-app-notifications.md
RytoEX: It sounds like you're saying that the downside is not "They will trigger a sound that the user cannot mute" but that "the user cannot customize which notifications provide a sound and which do not".

Notifications persist in the notification list up to the limit set in the notification settings (see image provided in [my comment on this PR](https://github.com/obsproject/rfcs/pull/19#issuecomment-633283181).

accepted/0019-app-notifications.md
```diff
@@ -0,0 +1,111 @@
...
+  - Upsides include:
+    - easier to implement
+    - will get the user's attention
+
+# Unresolved questions
+
+* Should Alerts have a list of methods to dismiss (scene collection switch, profile switch)?
```
DDRBoxman, line 107: I think the single queue with manual dismiss is more than adequate.

accepted/0019-app-notifications.md
```diff
@@ -0,0 +1,111 @@
...
+  * Methods to dismiss would be quite limited in libobs, and would add unneeded complexity to the frontend. Methods to dismiss would instead be tied to sources, outputs, etc.
+* The focus of this current design is of the assumption that alerts are only shown in the main window. Often it makes more sense to show warnings/info within Properties themselves. I personally think the scope should be limited to the core alert system, and overhauling Properties to support many more things is a task for another day. Thoughts?
```
DDRBoxman, line 109: Agreed, property alerts are out of scope for this RFC.

accepted/0019-app-notifications.md
```diff
@@ -0,0 +1,111 @@
...
+* Rather than making the summary/description link/anchor capable, should we instead provide a "link" field, and auto generate a "Learn more" link at the end of the summary? The big downside of this is then only one link can be provifded, rather than providing extensibility and further information for the user.
+* Should this Alert API contact an OBS server to recieve alerts? Use case here would be if a service is having connectivity issues that is out of our control. This could explicitly be optional, maybe only enabled when the user links their account with a service.
```
DDRBoxman, line 111: Probably also out of scope for the RFC, but it would be cool to see service plugins add support for the alert system after the fact.

DDRBoxman:
I believe this RFC adequately describes the proposed alerting system, I'm moving this RFC into FCP.

Fenrirthviti:
This has been in FCP for a while, with no objections, going to go ahead and merge.

PR #20, opened 2020-04-18
Update README.md to improve process

Fenrirthviti:
This updates the RFC process to something a bit more complete. Process has been taken from the Rust RFC repository, and modified to fit the OBS Project needs and requirements. This is open for discussion as parts of this process I need assistance and opinions on how to define, such as examples of a substantial change to OBS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ community.
 
 The "RFC" (request for comments) process is intended to provide a consistent
 and controlled path for new features, functionality, and changes to undergo
-a peer-review and allow all project stakeholders to weigh in their comments
+a peer-review and allow all interested parties to weigh in their comments
 and ensure the direction is consistent with the vison of the project.
 
 
@@ -40,7 +40,10 @@ OBS Studio, obs-browser, or any of the other plugins/modules. What constitutes a
 on what part of the ecosystem you are proposing to change, but may include the
 following.
 
-  - ????
+  - Entirely new features/functionality that are not currently available
+  - Major changes to existing features, or requests to redesign those features
+  - Things that drastically change the end-user experience, both from a
+    workflow and UX perspective.
 
 Some changes do not require an RFC:
 
@@ -119,14 +122,14 @@ It is recommended that no work start on implementation until an RFC is accepted.
     needs to have already been clearly articulated, and there should not be a
     strong consensus *against* that position outside of the team. Team
     members use their best judgment in taking this step, and the FCP itself
-    ensures there is ample time and notification for stakeholders to push back
-    if it is made prematurely.
+    ensures there is ample time and notification for interested parties to push
+    back if it is made prematurely.
     - For RFCs with lengthy discussion, the motion to FCP is usually preceded by
       a *summary comment* trying to lay out the current state of the discussion
       and major tradeoffs/points of disagreement.
   - The FCP lasts ten calendar days, so that it is open for at least 5 business
-    days. This way all stakeholders have a chance to lodge any final objections
-	before a decision is reached.
+    days. This way all parties have a chance to lodge any final objections
+    before a decision is reached.
   - In most cases, the FCP period is quiet, and the RFC is either merged or
     closed. However, sometimes substantial new arguments or ideas are raised,
     the FCP is canceled, and the RFC goes back into development mode.
@@ -137,8 +140,8 @@ It is recommended that no work start on implementation until an RFC is accepted.
 Once an RFC becomes "active" then authors may implement it and submit the
 feature as a pull request to the OBS Porject (or appropriate module) repo. Being 
 "active" is not a rubber stamp, and in particular still does not mean the feature
-will ultimately be merged; it does mean that in principle all the major stakeholders
-have agreed to the feature and are amenable to merging it.
+will ultimately be merged; it does mean that in principle all the major interested
+parties have agreed to the feature and are amenable to merging it.
 
 Furthermore, the fact that a given RFC has been accepted and is "active"
 implies nothing about what priority is assigned to its implementation, nor does
@@ -204,8 +207,9 @@ closed (as part of the rejection process). An RFC closed with "postponed" is
 marked as such because we want neither to think about evaluating the proposal
 nor about implementing the described feature until some time in the future, and
 we believe that we can afford to wait until then to do so. Postponed pull
-requests may be re-opened when the time is right. We don't have any formal
-process for that, you should ask members of the core team.
+requests may be re-opened when the time is right. There is no formal process
+for when a postponed RFC is reopened. We recommend asking us directly about
+any postponed RFCs in our [official Discord Server].
 
 Usually an RFC pull request marked as "postponed" has already passed an
 informal first round of evaluation, namely the round of "do we think we would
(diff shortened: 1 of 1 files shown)


DDRBoxman:
Excellent 🚀

README.md
```diff
@@ -1,19 +1,238 @@
...
+be demotivating for the unprepared contributor. Laying some groundwork ahead of
+the RFC can make the process smoother.
+
+Although there is no single way to prepare for submitting an RFC, it is
+generally a good idea to pursue feedback from other project developers
+beforehand, to ascertain that the RFC may be desirable; having a consistent
+impact on the project requires concerted effort toward consensus-building.
+
+The most common preparations for writing and submitting an RFC include talking
+the idea over on our [official Discord server], or discussing the topic on our
```
DDRBoxman, line 74: I guess these become links eventually?

DDRBoxman:
We probably want a bit in here about how RFCs should lay out technology choices and specific details related to implementation.

README.md
```diff
@@ -1,19 +1,238 @@
...
+## When you need to follow this process
+[When you need to follow this process]: #when-you-need-to-follow-this-process
+
+You need to follow this process if you intend to make "substantial" changes to
+OBS Studio, obs-browser, or any of the other plugins/modules. What constitutes a
+"substantial" change is evolving based on community norms and varies depending
+on what part of the ecosystem you are proposing to change, but may include the
+following.
+
+  - ????
```
dodgepong: I think "substantial" changes would either be entirely new features or changing the implementation of an existing feature/module.

dodgepong:
I wonder if we need a section somewhere explaining why we even have an RFC process. There's a section about how this process may seen too informal, but my suspicion is that many people will perceive this instead as too formal with too much red tape.

README.md
```diff
@@ -1,19 +1,238 @@
...
+    needs to have already been clearly articulated, and there should not be a
+    strong consensus *against* that position outside of the team. Team
+    members use their best judgment in taking this step, and the FCP itself
+    ensures there is ample time and notification for stakeholders to push back
+    if it is made prematurely.
+    - For RFCs with lengthy discussion, the motion to FCP is usually preceded by
+      a *summary comment* trying to lay out the current state of the discussion
+      and major tradeoffs/points of disagreement.
+  - The FCP lasts ten calendar days, so that it is open for at least 5 business
+    days. This way all stakeholders have a chance to lodge any final objections
```
dodgepong: I realize "stakeholders" is being used here colloquially, but I wonder if we should use a different word here anyway, such as "interested parties".

README.md
```diff
@@ -1,19 +1,238 @@
...
+## RFC Postponement
+[RFC Postponement]: #rfc-postponement
+
+Some RFC pull requests are tagged with the "postponed" label when they are
+closed (as part of the rejection process). An RFC closed with "postponed" is
+marked as such because we want neither to think about evaluating the proposal
+nor about implementing the described feature until some time in the future, and
+we believe that we can afford to wait until then to do so. Postponed pull
+requests may be re-opened when the time is right. We don't have any formal
+process for that, you should ask members of the core team.
```
dodgepong: The grammar of this sentence seems weird, not sure what it's trying to say. Should we have a formal process for this, or should we just instruct people to ping project members on the closed PR?

dodgepong:
A few minor comments but otherwise looks really good. Thanks for writing this up!

README.md
Fenrirthviti: I like this, I'll make the changes tonight. I left it in where it was being used loosely fro lack of a better term, but I think interested parties is more descriptive for the audience we're actually targeting.

README.md
Fenrirthviti: I don't think having a formal process serves much purpose since it boils down to "when we're ready". I think the intent is that we provide the reason and give a general time-frame in the FCP.

Fenrirthviti:
> We probably want a bit in here about how RFCs should lay out technology choices and specific details related to implementation.

@DDRBoxman Can you give me an example or rough draft for wording here, and a suggestion where that should be? I'm thinking under the `Before creating an RFC` section, a call out to take those things in to consideration.

Fenrirthviti:
Also, updated per @dodgepong's comments. Leaving the changes as separate commits for now, will squash them when it is time to merge.

README.md
N1GHTR4NG3R: Agreed, Substantial changes should reflect anything that replaces existing features, the GUI or implementations of things.

jp9000: This can be:
- Anything that adds a major feature
- Anything that changes internal engineering design of any part of the program.  For example, changing the slideshow to be dynamic rather than static would be considered a big engineering design change
- Anything that changes UX/UI design, or changes something that the user has to interact with to a significant extent and it's a big change.
- Anything that changes API, ABI, or settings storage format of the program

There probably can be exceptions to this.  For example, if it's a small change where your work isn't really "wasted" if it's rejected, then it's probably fine to directly PR without an RFC.  Another example is where a change is sort of obviously ideal, like making a pull request to add texture-based encoding for QSV, that is a recent fairly significant change but is sort of "obvious".  Or some things written by maintainers and/or trusted core contributors (although even they are encouraged to use RFCs).  Or things that have already been approved of by the maintainers, and discussion has already happened in some form somewhere else.

jp9000:
I left one additional comment, and also, the only thing I want to note is if there is a way to make this whole doc more concise, that would be helpful I think.  Human beings can just sort of zone out after a while.  So having tight, clear messaging with minimal repeated information should probably be key to get people to know what they've got to do without having checked out of reading before the end.

(I'm guilty of writing this way as well by the way)

Fenrirthviti:
@jp9000 I've updated with your comments on when to submit an RFC. I'm not sure we can make this any more concise without losing effectiveness. To be honest, it already feels too concise. This is our process, it should be well defined and not left to ambiguity (which many parts of it honestly already are). The more well defined this document is, the less issues we will have in the long run. I understand what you mean, but I respectfully disagree on cutting process documentation down to bare minimum.

dodgepong:
> I'm not sure we can make this any more concise without losing effectiveness.

Maybe we could try putting a TL;DR version of the process toward the top of the document?

Fenrirthviti:
I don't like that, because then it discourages reading the document. I get `humans are lazy they don't want to read, etc. etc`, but I don't feel like we should be catering to that mentality. If contributors don't want to take the 5 minutes to read our process, they frankly shouldn't be contributing to the project. Right now the process is very clear with the section titles, you can basically just skip to the "what the process is" and get the majority of what you need to do in the first 5 bullet points.

Again, I completely understand where you are coming from but this is not the OBS Studio application. We should not be encouraging or embracing that level of hand-holding for this.

Fenrirthviti:
Going to go ahead and merge this for now, as the current process document we have is very lacking. This is still considered open for review and comments, and we can certainly update if there are any strong objections, but I want to get this in front of folks submitting RFCs as the activity on this repo is starting to pick up significantly, and it seems like there are no strong objections to any of the actual process or content, just a minor disagreement on how much we should summarize. I am open to arguments or suggestions on how to best handle that, but I do feel pretty strongly that we shouldn't be encouraging tl;dr when it comes to developers.

PR #21, opened 2020-04-18
Add RFC: Flatpak (Linux)

GeorgesStavracas:
- Start Date: 2020-04-18
- RFC PR: #21
- Mantis Issue: N/A

# Summary

Flatpak is an app packaging and distribution mechanism widely available on Linux
distributions. To some distributions, Flatpak is part of the default experience.
Flatpak allows tighter control over the environment applications run in, how they
are packaged and installed, and also isolates the application from the host
system.

The goal is add support for Flatpak, both as a development platform, and also as
a distribution platform.

# Motivation

As a complex multi-platform project, OBS Studio has a wide surface for breakage.
Bad packaging can be a real problem, given that many Linux distributions package
OBS Studio in slightly different and incompatible ways.

By supporting Flatpak, OBS Studio benefits from having a tight control over the
execution environment, and the packaging of plugins and dependencies. This will
reduce the number of moving part when running OBS Studio, which allows much easier
reproduction of bugs and, consequently, fixing them.

## Internals

There are no code changes involved in adding Flatpak support. It is simply a
matter of adding a new file. This file is JSON formatted file containing the
dependencies, permissions, and the platform that OBS Studio depends on.

# How We Teach This

Because this is an addition to the platform, users shouldn't be required to learn
about Flatpak. For developers, little will change.

# Drawbacks

No known drawbacks.

# Additional Information

Supporting Flatpak does not prevent OBS Studio from supporting other distribution
mechanisms, nor will affect Linux distributions that package OBS Studio manually.

# Unresolved questions

 * Should OBS Studio use Flatpak as part of the CI process?
 * Should Flatpak be part of the release process?

### DIFF
--- a/accepted/0020-flatpak.md
+++ b/accepted/0020-flatpak.md
@@ -1,0 +1,50 @@
+- Start Date: 2020-04-18
+- RFC PR: #21
+- Mantis Issue: N/A
+
+# Summary
+
+Flatpak is an app packaging and distribution mechanism widely available on Linux
+distributions. To some distributions, Flatpak is part of the default experience.
+Flatpak allows tighter control over the environment applications run in, how they
+are packaged and installed, and also isolates the application from the host
+system.
+
+The goal is add support for Flatpak, both as a development platform, and also as
+a distribution platform.
+
+# Motivation
+
+As a complex multi-platform project, OBS Studio has a wide surface for breakage.
+Bad packaging can be a real problem, given that many Linux distributions package
+OBS Studio in slightly different and incompatible ways.
+
+By supporting Flatpak, OBS Studio benefits from having a tight control over the
+execution environment, and the packaging of plugins and dependencies. This will
+reduce the number of moving part when running OBS Studio, which allows much easier
+reproduction of bugs and, consequently, fixing them.
+
+## Internals
+
+There are no code changes involved in adding Flatpak support. It is simply a
+matter of adding a new file. This file is JSON formatted file containing the
+dependencies, permissions, and the platform that OBS Studio depends on.
+
+# How We Teach This
+
+Because this is an addition to the platform, users shouldn't be required to learn
+about Flatpak. For developers, little will change.
+
+# Drawbacks
+
+No known drawbacks.
+
+# Additional Information
+
+Supporting Flatpak does not prevent OBS Studio from supporting other distribution
+mechanisms, nor will affect Linux distributions that package OBS Studio manually.
+
+# Unresolved questions
+
+ * Should OBS Studio use Flatpak as part of the CI process?
+ * Should Flatpak be part of the release process?


kkartaltepe:
> Flatpak is an app packaging and distribution mechanism widely available on Linux
distributions. To some distributions, Flatpak is part of the default experience.

The same applies to literally every packager available in linux at the current time. Perhaps you can compare flatpak to existing options in terms of real tangible benefits instead of speaking in platitudes.

> Flatpak allows tighter control over the environment applications run in, how they
are packaged and installed, and also isolates the application from the host
system.

This is mostly opposed to the functions of OBS which is to be tightly integrated into the desktop environment (i.e. capturing it). Isolation typically works to defeat the purpose of OBS. So im not sure how this is a benefit.

As an example OBS typically relies on driver provided functionality like opengl and vaapi and nvenc. Vaapi is known to break compatibility with old drivers so trying to isolate the obs userland would only cause more breakages then relying on packagers.

As a second example downstream flatpak requires xdg-portal extension because normal x capture is presumably blocked in the flatpak runtime environment (maybe it happens to work).

> Should Flatpak be part of the release process?

I dont see any reason to add it unless we intend to support it as a releasable package. It would only bait people into thinking we support it if we add it otherwise.

GeorgesStavracas:
Moving the contents of https://github.com/obsproject/obs-studio/pull/2752 to here. [This is the proposed Flatpak manifest file](https://gist.github.com/GeorgesStavracas/55343416e2f79994693754d2c15dbac2).

I'm personally excited with the prospects of Flatpak support in regards to the development experience as well. Flatpak allows a "clone & run" approach, [pretty much like this](https://youtu.be/8WSkH6NK-A0).

kkartaltepe:
as an example of how isolation in flatpak itself is also misaligned with obs. Consider from the proposed manifest rule.

```
    "--socket=x11",
    "--socket=pulseaudio",
    "--device=all",
    "--share=network",
    "--share=ipc",
    "--filesystem=xdg-run/obs-xdg-portal:create",
    "--filesystem=host",
    "--talk-name=org.kde.StatusNotifierWatcher",
    "--talk-name=org.freedesktop.ScreenSaver",
    "--talk-name=org.freedesktop.PowerManagement.Inhibit",
    "--talk-name=org.freedesktop.Notifications",
    "--talk-name=org.mate.SessionManager",
    "--talk-name=org.gnome.SessionManager",
```

Who maintains these? How do we know what new features might require? Are we going to get a new fun "talk-name" to add for ever DE someone can think up? What about Jack (looking at pulseaudio). Does ALSA bypass this?

If the only permissions to add were those maintained by flatpak and freedesktop it might seem almost reasonable. However it obviously depends on per DE magic which is not appealing.

And of course in its goal of isolation flatpak obfuscates the actual DE being run, preventing logging for providing useful information (since we have established that flatpak is dependent on DEs). e.g. obs will log `info: Distribution: "KDE Flatpak runtime" "5.14"` or such.

fastoslinux:
it would be wonderful to have OBS Studio on Flathub officially, the current version works very well!

dodgepong:
As I understand it, the current build on Flathub has some third-party plugins bundled with it, such as obs-linuxbrowser rather than obs-browser. Apparently there are some features of obs-linuxbrowser that aren't on obs-browser yet, such as the ability to load CSS from an external file. If that transition is to be made nicely, it would be nice for those features to be submitted to obs-browser.

GeorgesStavracas:
> it would be wonderful to have OBS Studio on Flathub officially, the current version works very well!

When I proposed this RFC, I wasn't necessarily thinking about publishing OBS Studio on Flathub directly, but that would be wonderful indeed. What I had in mind is just adding a Flatpak manifest file to OBS Studio itself, which would then go through a period of more extensive testing, before we can think of publishing it somewhere.

> As I understand it, the current build on Flathub has some third-party plugins bundled with it, such as obs-linuxbrowser rather than obs-browser. Apparently there are some features of obs-linuxbrowser that aren't on obs-browser yet, such as the ability to load CSS from an external file. If that transition is to be made nicely, it would be nice for those features to be submitted to obs-browser.

Those are good points, I absolutely agree that the Flathub build shouldn't ship plugins that are not default. I'm not proud that one of these out-of-tree plugins is my own `obs-xdg-portal`, which is something I am currently working on moving to OBS Studio itself. From recent Discord conversations, it seems like it is a reasonable plugin to upstream.

As for the `obs-linuxbrowser` plugin, while I'm a slightly scared to make a commitment to it, I can probably put it in my queue for investigating, discussing, and potentially upstreaming of its features into `obs-browser`.

I believe the moment Flathub doesn't need to ship any out-of-plugin by default, is the moment OBS Studio should consider publishing to it (if desired, of course). But I don't personally think it blocks us from having a Flatpak manifest available before that.

GeorgesStavracas:
I'd like to take a step back and address @kkartaltepe's concerns. Sorry this is a wall of text, but I hope to shed some light on the points you raised.

> Perhaps you can compare flatpak to existing options in terms of real tangible benefits instead of speaking in platitudes.

As a developer the sandboxing side of Flathub aimed at reducing the gap between "works on my machine" and "works on anyone else's machine". As such it tries to promote specific, reusable and reproducible builds and a common runtime, whilst still being "desktop oriented" with fairly seamless integration for video drivers, portals, acceptable desktop permissions etc.

My intention is not to advocate for Flatpak in exclusion of other technologies. but there are some good reasons for having a Flatpak manifest available:

 * The manifest can be built by any distribution, and does not depend on the distribution to work. The generated package can be installed by any distribution, regardless of their own package manager. An OBS Studio Flatpak can be installed on Debian, Ubuntu, Fedora, SilverBlue, Endless OS, etc, without any changes. (In fact, you can find developers of all these distributions, and others, contributing to Flatpak.)
 * This "one package everywhere" approach reduces *a lot* the number of variables you have to deal with. On the development side, if someone reports a bug, we can ask the reporter to test with the Flatpak version, and iron out bugs that are caused by mispackaging distros. This is why you want this integrated with CI so you get a single testable artifact.
 * A Flatpak manifest also makes it clear what the supported and expected libraries are to distributions (it can be as specific as exact releases) both in terms of how to compile and what is the expected environment to run the application. The intention is that over time you don't end up with the multiple different ways to compile that you see on the 'how to build' page today.

> This is mostly opposed to the functions of OBS which is to be tightly integrated into the desktop environment (i.e. capturing it). Isolation typically works to defeat the purpose of OBS.

A properly sandboxed application **requires** tight integration to the platform, by using the appropriate screencasting APIs. In this case, as discussed on Discord, the best API we have for now is the [XDG Desktop portal](https://github.com/flatpak/xdg-desktop-portal). Specifically, the [ScreenCast interface](https://github.com/flatpak/xdg-desktop-portal/blob/master/data/org.freedesktop.impl.portal.ScreenCast.xml).

By using these APIs, OBS Studio will work inside and outside sandboxes (be it Flatpak, Snap, or AppImage), on Wayland and X11, on all desktops that implement it. Right now, the ScreenCast interface is implemented by GNOME, KDE, and wlroots, which covers a fair share of desktop users.

Desktops that don't implement this interface can always fallback to good old X11 which works without change.

> As a second example downstream flatpak requires xdg-portal extension because normal x capture is presumably blocked in the flatpak runtime environment (maybe it happens to work).

The `obs-xdg-portal` plugin is there for Wayland support, and only that. The X socket is still readable the the xcomposite plugin works fine. With time, and if the OBS community accepts, I am willing to move the functionality of that plugin to `linux-capture`.

> Who maintains these?

If you're talking about the permissions, it's OBS Studio. We control which permissions we require in order to function correctly.

> Are we going to get a new fun "talk-name" to add for ever DE someone can think up?

I don't think so. The session management D-Bus specifically is a mess, as the result of a historical process. [But it's a controlled mess](https://gitlab.gnome.org/GNOME/gtk/-/blob/master/gtk/gtkapplication-dbus.c#L33). OBS Studio already support the standard and legacy D-Bus interfaces, and doesn't need to support more than that.

> What about Jack (looking at pulseaudio). Does ALSA bypass this?

Jack is an open question, but there [is rudimentary support for it](https://github.com/flatpak/flatpak/issues/1509). Admittedly, I just grabbed Flathub's manifest, and stripped out of what's unnecessary. If you look at the bottom of the file, you can appreciate the following lines:

```json
      "config-opts": [
        "-DCMAKE_BUILD_TYPE=Release",
        "-DUNIX_STRUCTURE=ON",
        "-DUSE_XDG=ON",
        "-DDISABLE_ALSA=ON",
        "-DDISABLE_JACK=ON",
        "-DENABLE_PULSEAUDIO=ON",
        "-DENABLE_WAYLAND=ON",
        "-DWITH_RTMPS=ON"
      ],
```
That is, when building as a Flatpak, OBS Studio explicitly disables Jack and ALSA. There are ongoing discussions around seamlessly enabling these in Flatpak, with ALSA more likely in the very near term. Jack is likely to be supported with pipewire compatibility.

> If the only permissions to add were those maintained by flatpak and freedesktop it might seem almost reasonable. However it obviously depends on per DE magic which is not appealing.

These permissions aren't "maintained by flatpak and freedesktop", we may have a misunderstanding. The permissions are the application defining the surface area that it needs and is willing to support. Or to put it another way, this moves the implicit contract into something that's explicit.

There are certainly some reasons why Flatpak may not be an ideal fit for the default distribution method for OBS Studio, whilst the Mesa, nvenc etc. story is pretty mature there are certainly some weakness' in Jack and ALSA support currently. These are areas that are actively being worked on. However for the majority of Linux cases this can be helpful in reducing the frustration with distributed development.

Fenrirthviti:
We also have several people working on an AppImage package. What advantages does this offer over AppImage? I believe this might be part of what @kkartaltepe was asking. What makes Flapak better than any of the other options? Why should we pick this one? Your comments seem to be addressing what the benefits of `anysimilarpackage` rather than addressing what makes Flatpak specifically better than something else. Every solution claims the same benefits, but if they were actually all the same, why would there be 4+ competing solutions?

That said, to be clear, I have no strong opinion on any of these solutions. I think having a platform-agnostic package that can be installed in any distribution is a positive thing for the project, but I don't want us running down multiple roads and incurring a heavy maintenance cost for what is effectively the same thing.

I'm afraid this might turn in to a personal opinion debate when it comes to Flatpak vs anything else, but I am curious to hear the opinions regardless.

l30br:
Flatpak and other forms of packaging are on the rise. I found this idea very positive.
I am carrying everything possible for flatpack and would be happy to have OBS too.

Moslley:
Flatpak and other forms of packaging are on the rise. I found this idea very positive.
I am carrying everything possible for flatpack and would be happy to have OBS too.

GeorgesStavracas:
Well, I was trying to avoid this kind of comparison. I don't feel comfortable in talking about one technology in detriment of others. Thus allow me to fully embrace that **this comment is very much my own personal view of these technologies**.

I don't think AppImage is really a solution to the problem at hand. AppImage is not an app distribution mechanism, it's an app _bundling_ mechanism. Once you bundle an AppImage with a set of libraries, the bundle is stuck with that forever. If that doesn't sound too bad, let me remind everyone of the existence of a [nightmare](https://github.com/intel/libva/issues) called VAAPI. Flatpak doesn't bundle anything to the application's executable, and it can do that thanks to runtimes. These low-level libraries are provided by the runtimes, and there is a team of volunteers and paid developers maintaining it. It's less the developer has to maintain and less security concern for the end user.

I mentioned Flatpak runtimes, and this is a key aspect of it. AppImage doesn't really have the concept of a runtime (i.e. "what does this app needs in order to run"), so you have to literally boot your computer with the oldest system you want to support, and build the AppImage inside it. This is not a joke, this is a [documented best practice](https://docs.appimage.org/reference/best-practices.html#general-recommendations) (!). Flatpaks don't need any of this because it addresses the true problem behind it: it isolates the application from the host system **for real**, at glibc level. It is, by design, guaranteed to work everywhere.

With Flatpak, updates are a fundamental part of the technology. If the app doesn't self-update, that's ok; if the app self-updates, that's ok too, Flatpak supports both. AppImage doesn't have automatic app updates, and if OBS Studio wants that on Linux, it'll have to write its own update mechanism. One could argue that you _can_ have AppImage updates, and that's true as long as you add a bunch of extra stuff to your host system. AppImage is _factually_ not distro-agnostic, it just works well enough on many of them.

Flatpak has deep desktop integration by default. Installing a Flatpak app shows it in the app list, it can export D-Bus interfaces, it has bash completion; basically, everything you expect of an app. This is something that AppImage doesn't have. AppImage wasn't built with that in mind. Again, if you really want to, you can, but you need to add a bunch of extra stuff to your system (but it has [quite considerable drawbacks](https://github.com/AppImage/AppImageKit/issues/58#issuecomment-388980893)). You have multiple choices on how you want your desktop applications to be half integrated into your desktop, pick your poison.

Flatpak is far easier to build because it doesn't ask require you to modify your program, or to [use any specific version of any fundamental library](https://docs.appimage.org/reference/best-practices.html#libstdc-so-6), or even to [binary-path the application](https://docs.appimage.org/reference/best-practices.html#binaries-must-not-use-compiled-in-absolute-paths). Flatpak doesn't force the application to statically link to anything. `flatpak-builder` makes building in a reproducible way far easier than AppImage where you usually just glue together your system libraries and hope it works.

At last, I'd like to mention that Flatpak can improve CI too. CI won't need to build and test OBS Studio on multiple distros, or pick one distro and run on it. By having an official, or even semi-official, distribution platform (for Flatpak, that's Flathub), OBS Studio can also avoid people running different versions built in different ways on [old](https://github.com/obsproject/obs-studio/issues/2747) or [exotic](https://github.com/obsproject/obs-studio/issues/2769) distros that have [incompatible](https://github.com/obsproject/obs-studio/issues/2644) [or](https://github.com/obsproject/obs-studio/issues/2654) [unreproducible](https://github.com/obsproject/obs-studio/issues/2645) [environments](https://github.com/obsproject/obs-studio/issues/2585).

refi64:
To add some comments to the post above: I *used* to be a huge fan of AppImages, but I ended up dropping them in favor of Flatpaks for the following reasons:

- With AppImages you have to manage updates by hand (i.e. have the software check for updates itself and instruct the user to download the file), whereas with Flatpak this is all handled for you.
- You have to manually send out AppImage updates if any of the likely many bundled dependencies are updated. Flatpak's runtimes store most of the deps apps would need, so the stuff you have to manually manage is heavily reduced.
- The only way to get true host integration is if the client system is running [appimaged](https://github.com/AppImage/appimaged) which:
  - is a bit of a hack (it scans download directories automatically for AppImages and pulls out the desktop files from inside, this turned me off big time from the project)
  - requires a daemon to be installed on the system, which removes one of the arguable advantages for AppImages (nothing would need to be present on the client side)
- glibc is not embedded, so you have to build on a distro with a relatively old glibc version, that way the software runs on all newer distros. This is pretty annoying if you rely on functionality that's hard to get in those older distros. Apparently the long term proposal for this is going to be to use a custom dynamic linker which...does not sound pretty (as someone who has had to mess with dynamic linkers in the past).
- The above also means that AppImages do not run on musl-based distros (Alpine, some Void variants).
- flatpak-builder makes it easier to have a reproducible build environment, in that anyone can run flatpak-builder and have it work. With AppImages, you'll often have to turn to e.g. Docker and work with containers manually in order to have some sort of reproducible build environment, which is a must because...
- ...It's easy to forget some libraries, so that the AppImage won't run on systems without those. Iirc this is supposed to be easily automated, but I would often run into weird edge cases that were frustrating.
- Related to the above, managing these reproducible environments while also mixing in CI isn't a very fun task.
- Flatpak has built-in systems for plugins to extend apps. These are all built in the same reproducible environments thanks to flatpak-builder. With AppImages, this is a bit hairier to handle; you'd generally either have your users save plugins to special directories and hope the versions stay in sync, or write your own plugin manager.
- The user has to mark AppImages as executable by hand, which isn't necessarily the best UX.
- Flatpak's sandboxing support is always a plus!

This isn't to say it's perfect (though nothing is), but I've definitely had *far* less trouble both building and managing Flatpaks instead, and even the most complicated stuff I've had to do generally didn't take as long as trying to make AppImages work consistently.

sudo-give-me-coffee:
@GeorgesStavracas you said a lot of wrong things:

> Flatpak doesn't bundle anything to the application's executable, and it can do that thanks to runtimes.

In Flatpak, an Application is linked to runtime until ap updates, is exactly same thing but with files that application don't need

> I mentioned Flatpak runtimes, and this is a key aspect of it. AppImage doesn't really have the concept of a runtime (i.e. "what does this app needs in order to run"), so you have to literally boot your computer with the oldest system you want to support, and build the AppImage inside it. This is not a joke, this is a documented best practice (!).

You not understand the concept, since glibc has retrocompatibility, once time you bundle the application on oldest base system, you don't need bundle glibc, do this will save up to 8 - 15 MB on resulting AppImage. The most troubles with AppImages is because its was build on newer system but doesn't bundle glibc

> It is, by design, guaranteed to work everywhere.

Not exactly, for e.g if for some reason `/tmp` has a 0 UUID owned files Flatpak fails to run, since Flatpak frequently breaks retro-compatibility with older versions. If  the distro doens't provide updated version of flatpak package frequently (such as Debian, Deepin) the user will not able to install or update newer applications and if try will no run applications anymore.  Of course is only some examples

> AppImage doesn't have automatic app updates

Wrong again, AppImages can easy self update just pass provide a .zsync file on HTTPS server and call zsync under AppDir, just chdir do AppImage location, for e.g. reading "APPIMAGE" environment variable, is very easy to do, 

> One could argue that you can have AppImage updates, and that's true as long as you add a bunch of extra stuff to your host system.

This is needed by Flatpak, no in AppImage, an external updates requires a single file, to install and update Flatpak, the user will need 104 files according Debian:

* [The Flatpak](https://packages.debian.org/buster/amd64/flatpak/filelist)
* [The libostree](https://packages.debian.org/buster/amd64/libostree-1-1/filelist)
* [The json-glib](https://packages.debian.org/buster/amd64/libjson-glib-1.0-0/filelist)

To update AppImages just a **single** file:

* [AppImageUpdate](https://github.com/AppImage/AppImageUpdate/releases/download/continuous/AppImageUpdate-x86_64.AppImage)

> AppImage is factually not distro-agnostic

Flatpak also

> Flatpak has deep desktop integration by default. Installing a Flatpak app shows it in the app list, it can export D-Bus interfaces, it has bash completion; basically, everything you expect of an app.

ApImages also export D-Bus interfaces, and **[if developer wants](https://github.com/aferrero2707/gimp-appimage/releases)** shows it in the app list after first run (similar to Windows, Mac OS X, Android...), Flatpaks doesn't have a lot of visual integrations, is a significant drawback especially media content apps, Qt Applications has a messy design outside LXQT and GTK apps looks like a alien outside GNOME OSes (Endless OS, Fedora Silverblue...) without work arounds

barthalion:
> In Flatpak, an Application is linked to runtime until ap updates, is exactly same thing but with files that application don't need

You're mistaken here. Application ships only libraries which are outside the runtime. Runtime is updated separately, so maintainer doesn't need to worry about core libraries such as graphical toolkit or SSL.

> You not understand the concept, since glibc has retrocompatibility, once time you bundle the application on oldest base system, you don't need bundle glibc, do this will save up to 8 - 15 MB on resulting AppImage. The most troubles with AppImages is because its was build on newer system but doesn't bundle glibc

You have proved yourself it's not portable, not to mention it implies being bound to a distribution shipping ancient software, with everything else that comes included. Old GCC, old curl, old ffmpeg. I'm also happy to notice that you conveniently skipped the part about musl distributions.

> Not exactly, for e.g if for some reason `/tmp` has a 0 UUID owned files Flatpak fails to run, since Flatpak frequently breaks retro-compatibility with older versions.

`[citation needed]`

> If the distro doens't provide updated version of flatpak package frequently (such as Debian, Deepin) the user will not able to install or update newer applications and if try will no run applications anymore. Of course is only some examples

Both Debian and Deepin are perfectly fine with flatpak versions shipped in their repositories. There are newer releases available in backports but it just means missing *optional* new features.

> Wrong again, AppImages can easy self update just pass provide a .zsync file on HTTPS server and call zsync under AppDir, just chdir do AppImage location, for e.g. reading "APPIMAGE" environment variable, is very easy to do,

It's very simple! Just serve a file from a server, chdir, call zsync, set proper environment variable and you are done! It sounds just ridiculous when compared to the integrated experience Flatpak provides with Flathub, completely ignoring integrated CI and CDN.

> This is needed by Flatpak, no in AppImage, an external updates requires a single file, to install and update Flatpak, the user will need 104 files according Debian:
> (snip)
> To update AppImages just a **single** file:
> 
>     * [AppImageUpdate](https://github.com/AppImage/AppImageUpdate/releases/download/continuous/AppImageUpdate-x86_64.AppImage)

Yes, these are distribution maintained packages providing core functionality of Flatpak, including support and security fixes for time frame decided by Debian. I'll take it any day over a random binary which promises to do the correct thing.

> ApImages also export D-Bus interfaces, and **[if developer wants](https://github.com/aferrero2707/gimp-appimage/releases)** shows it in the app list after first run (similar to Windows, Mac OS X, Android...), Flatpaks doesn't have a lot of visual integrations, is a significant drawback especially media content apps, Qt Applications has a messy design outside LXQT and GTK apps looks like a alien outside GNOME OSes (Endless OS, Fedora Silverblue...) without work arounds

Have you ever actually used Flatpak or just spreading FUD? After initial setup (which requires relogin for distributions which don't come with it pre-installed), desktop launchers are visible instantly after an application is installed. Theming is tricky to get right in a secure way, so Flatpak will instead look for the specific theme in the repository and install it as an extension.

In general, you are comparing apples to oranges. Even Microsoft realized that run-this-random-exe-from-internet is a flawed idea and added an official application store with Windows 10 and started working on sandboxing applications.

sudo-give-me-coffee:
> You're mistaken here. Application ships only libraries which are outside the runtime. Runtime is updated separately, so maintainer doesn't need to worry about core libraries such as graphical toolkit or SSL.

Nope, apps is separated from runtime, but still linked by this, so if a application depends on org.kde.Platform version 5.14 they will run with these libs until an application rebuild with a different version/runtime, the only difference is that on AppImage application files and runtime files is separated

> You have proved yourself it's not portable, not to mention it implies being bound to a distribution shipping ancient software, with everything else that comes included. 

You can also bundle newer libs and run on old systems. But if the oldest version works perfectly why uses newer? This generally implies in more resource usage for things that is never used

> I'm also happy to notice that you conveniently skipped the part about musl distributions.

If glibc is bundled, AppImage runs normally

> [citation needed]

https://github.com/flatpak/flatpak/issues/3470
https://github.com/flatpak/flatpak/issues/2948
https://github.com/containers/bubblewrap/issues/273 (apparently related)

Of course you can test by yourself

> Both Debian and Deepin are perfectly fine with flatpak versions shipped in their repositories.

**NEWER** versions (Debian Buster and Deepin 20 Beta), but during some time these distros was losed support, and backports isn't enabled by default

> ! It sounds just ridiculous when compared to the integrated experience Flatpak provides with Flathub, completely ignoring integrated CI and CDN.

Do it one time. And the point was "AppImage doesn't have automatic app updates", this was wrong

> Yes, these are distribution maintained packages providing core functionality of Flatpak, including support and security fixes for time frame decided by Debian.

Please conclude your argument

> I'll take it any day over a random binary which promises to do the correct thing.

Flatpak is exactly it

> desktop launchers are visible instantly after an application is installed

Only works seamless under GNOME, by default is needed relogin **for each app** to show on menu, tested:

* Elementary OS 6.1
* Endless OS > Works
* Fedora Silverblue (GNOME and KDE) > Works on Gnome
* OpenSuse > Works on Gnome
* Debian Buster > Works on Gnome
* All Ubuntu flavours (18.04-20.04 beta) > Works on main and MATE

> Theming is tricky to get right in a secure way, so Flatpak will instead look for the specific theme in the repository and install it as an extension.

Since it is read only, it doesn't make any sense, this is a limitation

sudo-give-me-coffee:
Are you serious that you can't say good things about Flatpak without passing incorrect information in another format? You can say, for e.g:

- Flatpaks can run on musl based distros without any extra addition

sudo-give-me-coffee:
The main trouble of Flatpak isn't with Flatpak itself, but on arguments used by who defends it as the best way to distribute applications

SISheogorath:
:eyes: Looking at the current discussion, I think this is becoming ridiculous. There are @refi64 and @barthalion, who are both involved in Flatpak and doing really great work in developing Flatpak and Flathub on one side and @sudo-give-me-coffee on the other side who has his/her strong preference for appimages. But for both sides I have to say: I don't think this is the right fighting ground to figure out if the univeral truth is to use flatpak or appimage. Technically speaking, I would say both have their reason to exist and there is no clear winner for all users out there. So maybe writing books of arguments and accusing each other of stuff is not the right way to go about this at all.

The decision has to be made by those who are willing to invest their time into maintaining these ways of software packaging. So maybe let the OBS developers decide?

What I can tell so far is that I'm a heavy flatpak user and have the OBS version from FlatHub installed. It worked great out of the box, I was able to record all windows I wanted to record, GNOME did a proper indication that the screen was recorded and that basically means that the integration works as intended.

Maybe that's something the developers can work with and I really hope, that this crusade can stop in here? How old are we?

kkartaltepe:
This is getting a little heated. There are some interesting points worth digging into. But lets not devolve to religious wars here.

And please note this is not a page to figure out which platform is the one true platform. Simply to see how well they align with OBS and our goals.

GeorgesStavracas:
This was asked on Discord, and I thought it would be an interesting data point to share here as well. Here are the statistics from Flathub for com.obsproject.Studio:

daily downloads

It mixes from-scratch and delta downloads, so it's a bit hard to analyse it without doing some guess work, but the spikes seem to indicate that there are at least 6,000 computers installing and updating OBS Studio from Flathub in a single day.

sudo-give-me-coffee:
@SISheogorath 

The only reason was the wrong information on https://github.com/obsproject/rfcs/pull/21#issuecomment-618080932 about AppImages

-------------------------------------------

They must Defend Flatpak without giving wrong information about another technology. If the incorrect information was about snaps, I would have done the same, even though I don't like the idea of ​​snaps

fastoslinux:
Most software centers are integrated (LinuxMint-mintinstall, Manjaro-pamac, GNOME Software, Discover on KDE, PopOS-appcenter....)

Captura de tela de 2020-04-23 14-11-42

In the Steam from Flathub example, extensions (for OBS would be third-party plugins) for the user to conveniently add:

Captura de tela de 2020-04-23 14-05-12

eszlari:
> > desktop launchers are visible instantly after an application is installed
> 
> Only works seamless under GNOME, by default is needed relogin **for each app** to show on menu

As a KDE Plasma user I can assure you, that it works perfectly there too!

dodgepong:
The reality here is that we don't have to choose between AppImage and Flatpak, we can do both.

Also, the more of this can be automated in terms of building and making updates, the better. That reduces the number of things that might break or have to be upgraded between updates, and makes it so that people with less special knowledge can maintain it if need be.

GeorgesStavracas:
I was intrigued by @kkartaltepe's first comments (thanks for pointing those issues out!) and quickly investigated if it would be possible to build OBS Studio on Flatpak with support to Jack and ALSA. Turns out it's possible! [Here's the updated manifest file](https://gist.github.com/GeorgesStavracas/55343416e2f79994693754d2c15dbac2).

The usual rules apply: you need to have a functional Jack server in order to use the Jack plugin.

hfiguiere:
And it needs flatpak 1.6.1 at least, at run time, to work.

kkartaltepe:
We'll lets test it out. Im unable to open file dialogs for the LUT video filter applied to a v4l2 source. Is this also happening on anyone elses flatpak (both the build instructions from the PR and the flathub package have this issue for me). building off the PR commit for both my local flatpak build and a non-flatpak version with no such issue.

dodgepong:
> We'll lets test it out

So, a bit of a philosophical question here: This is an RFC, so that typically implies that the implementation hasn't been done. Obviously, we can see a proof-of-concept already, but really the RFC discussion is intended to answer the question of "should we do this, and if so, how should it be done?" So in that sense, testing the POC might be outside the scope of the RFC, since the question is more about design than implementation quality.

On the other hand, in order to make a determination of whether or not this will work for us, it's hard to say without testing _something_. Is it appropriate to test and iterate on a POC during an RFC?

kkartaltepe:
>  Is it appropriate to test and iterate on a POC during an RFC?

I dont see why not? In particular the claims in the RFC are that it should aid in replicating issues. So if we have an issue i think seeing if it replicates should help bolster the argument made in the RFC.

Personally I feel the debugability is an important aspect in any packaging. With native packaging its trivial, we built it there and we can test it there and use any tools we want. In particular packaging other than native is an added layer that must be debuggable (and this issue seems directly related to the packaging or runtime), so what it takes to resolve an issue seems on topic to me.

GeorgesStavracas:
> Im unable to open file dialogs for the LUT video filter applied to a v4l2 source. Is this also happening on anyone elses flatpak

I can confirm that. It doesn't open the browse dialog and throws `(obs:2): Gtk-DEBUG: 09:26:10.314: Failed to create D-Bus proxy: Could not connect: No such file or directory` at the terminal. Can you confirm this message?

Captura de tela de 2020-04-24 09-27-13

kkartaltepe:
I have no such debug messages.

eszlari:
Under KDE Plasma I get:
```
org.kde.solid.udisks2: Failed enumerating UDisks2 objects: "org.freedesktop.DBus.Error.Disconnected" 
 "Not connected to D-Bus server"
org.kde.solid.udisks2: Failed enumerating UDisks2 objects: "org.freedesktop.DBus.Error.Disconnected" 
 "Not connected to D-Bus server"
org.kde.solid.udisks2: Failed enumerating UDisks2 objects: "org.freedesktop.DBus.Error.Disconnected" 
 "Not connected to D-Bus server"
org.kde.solid.udisks2: Failed enumerating UDisks2 objects: "org.freedesktop.DBus.Error.Disconnected" 
 "Not connected to D-Bus server"
org.kde.solid.udisks2: Failed enumerating UDisks2 objects: "org.freedesktop.DBus.Error.Disconnected" 
 "Not connected to D-Bus server"
org.kde.solid.udisks2: Failed enumerating UDisks2 objects: "org.freedesktop.DBus.Error.Disconnected" 
 "Not connected to D-Bus server"
org.kde.solid.udisks2: Failed enumerating UDisks2 objects: "org.freedesktop.DBus.Error.Disconnected" 
 "Not connected to D-Bus server"
org.kde.solid.udisks2: Failed enumerating UDisks2 objects: "org.freedesktop.DBus.Error.Disconnected" 
 "Not connected to D-Bus server"
```

eszlari:
It's a bug in Qt's portal code.

1. `flatpak run --command=sh com.obsproject.Studio`
2. `rm /run/user/1000/flatpak-info` (file is used to detect it's running inside flatpak)
3. `obs` (dialog shows up)

@grulja maintains that code.

grulja:
I'm not sure it's a bug in Qt (or my implementation). The reason why it fails is that you seem to set wrong filter, or something that Qt is not able to work with correctly. The QFileDialog::setNameFilter() should either be in form of `User label (*.png)` or `*.png`. Based on what it parses and sends through the portal, it looks like it's being used as `(*.png)`, which means the user label is missing. The non-portal QFileDialog handles this the way that it doesn't show any filter, at least when I tried non-flatpak OBS, it allows me to choose only *.png files, but the filter field in the file dialog is empty. For the flatpak version, this is parsed and send over DBus, where the user visible string is empty and the rest is fine. It looks like this:
```
      dict entry(
         string "filters"
         variant             array [
               struct {
                  string ""
                  array [
                     struct {
                        uint32 0
                        string "*.cube"
                     }
                     struct {
                        uint32 0
                        string "*.png"
                     }
                  ]
               }
            ]
      )

```
From Qt side it's fine, but this gets rejected by xdg-desktop-portal:
```
error time=1587751011.241648 sender=:1.117 -> destination=:1.594 error_name=org.freedesktop.portal.Error.InvalidArgument reply_serial=72
   string "invalid filter: name is empty"
```

GeorgesStavracas:
Thanks @grulja, your assessment looks correct! The above PR (https://github.com/obsproject/obs-studio/pull/2797) makes the dialog show properly now when running as a Flatpak.

kkartaltepe:
The proposed jack solution is precisely what the jack developer warns us not to do. We dont want to become the mispackaging distro. https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1001#note_323948324 (the pipewire developer is also not very optimistic about this conversation it seems.).

sudo-give-me-coffee:
Since JACK libs must not be bundled, you will need:

- [ ] a way to provide access to real jack libs, (e.g. creating a runtime containing the jack libs and their dependencies on install time)
- [x] allow access to jack shared memory dirs, (apparently is not standardized)
- [ ] handle the runtime containing jacklibs for when system updates
- [ ]  of course build the runtime with latest glibc everytime and update the application to be compatible with newer glibc libs

This will increase the chance of success, but no guaranteed to work

I has based on this:

> This is not possible. We cannot know if libjack.so from the system is compatible with the runtime in Flatpak. For example, if you have a very fresh distribution with a newer glibc than the runtime an application uses, then it might break. Even just the abi from GCC. For example if libssp is enabled on host. - Valentin David


Source: last comments in https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1001

kkartaltepe:
flatpak doesnt appear to support multi display X setups. At least on my system i couldnt figure out how to capture display :1. Maybe like before some member of the flatpak runtime team can enlighten us.

My flatpak obs install continues to corrupt the filesystem it resides on making using it increasingly annoying to use. Currently there are at least 3 duplicates of the plugin files within the runtime and this number appears to only go up as i relaunch obs. I thought this was from running the provided install commands multiple times but I have only installed once since uninstalling it and yet the libraries duplicated.

alatiera:
@kkartaltepe can you try running `flatpak repair` (like need to be prefixed with sudo) and see if it complains about anything? Else you can drop by #flatpak on Freenode or open an issue so we can figure out what's wrong with it.

GeorgesStavracas:
Sorry for the lack of response here, it's been post-release crunch on day job.

I've made some research to understand the problem domain, and seems like PipeWire is a good medium-to-long-term solution to bridge JACK support.

On Flatpak side, the next [FreeDesktop platform](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/2427) will ship a `libjack.so` provided by `pipewire-jack`. I couldn't find a nightly or master version of "org.kde.Platform" (if someone knows about it please let me know!) so we'll either have to ship `pipewire-jack` ourselves, or wait until the KDE platform receives it.

This `libjack.so` library aims to be a drop-in replacement of JACK but uses PipeWire internally. As far as I can tell, it effectively cuts this hard ABI dependency that the JACK daemon and JACK clients have on the library. With Flatpak, there's another handy factor: platforms are guaranteed to not break ABI.

Unfortunately, all of these points are PipeWire 0.3 features. I suppose supporting PipeWire is a medium-term bet, it'll probably pay off during the next year or so. Now, asking more conservative distros, like Ubuntu LTS or Debian Stable or Red Hat Enterprise Linux, to ship PipeWire 0.3 is going to be a hard sell.

I've asked Debian folks and it seems like this can be achieved with "backports", which I guess it's a repository with packages that will be part of the next Debian Stable version. Ubuntu seems to be streamlined with this. Looks like we can't expect it from RHEL though.

TiZ-HugLife:
Hello, I'd like to contribute to the discussion from an end user standpoint. I like to use stable distros like Xubuntu (considering Debian too), but I don't trust that they'll have up-to-date versions of applications like OBS in their repos. I don't like to have many PPAs installed because they make release upgrades a pain, and removing them to return the system to a sort of distro-kosher state doesn't always completely work out.

I used to build OBS from source, because I wanted the websocket plugin to control it with scripts, and I wanted to be sure I could encode with my NVidia GPU.

It turns out that the current version of OBS in Flatpak encodes with my NVidia GPU just fine, and the debian package for obs-websocket can simply be unpacked with a modified directory layout in `$XDG_CONFIG_HOME/obs-studio/plugins` and works as-is.

I trust Flathub and Flatpak to give me up-to-date applications to run on a stable desktop base, and as a way to run proprietary applications like Zoom and MS Teams without polluting my package manager or overexposing my system to them. (I know OBS is FOSS, just citing one of the advantages I find.)

As an end user, I find great value in Flatpak and would like to encourage upstream support.

azubieta:
Regarding the issue with the AppImage format, I would like to invite you to check my PR for packaging obs as an AppImage: https://github.com/obsproject/obs-studio/actions/runs/105687432

It ~~supports VAAPI~~, embeds glibc, and it's build on a modern system. Updates are not configured yet but they can be easily added.

To manage system integration and updates you can use https://github.com/AppImageCrafters/appimage-manager

Edit
===
Notice that a new AppImage packaging tool is used to create this bundle: [appimage-builder](https://appimage-builder.readthedocs.io/en/latest/). This has several advantages over the old tools, I would like to invite you to read the docs linked above.

derrod:
> It supports VAAPI

It doesn't on my notebook (Ubuntu 20.04, Intel i7 with QSV and NVidia GPU with NVENC). It just throws some errors. Natively installed OBS from the PPA works fine though.

azubieta:
> > It supports VAAPI
> 
> It doesn't on my notebook (Ubuntu 20.04, Intel i7 with QSV and NVidia GPU with NVENC). It just throws some errors. Natively installed OBS from the PPA works fine though.

Ok, I had reports saying that it's supported. I don't have a computer with an NVidia card at hand right now. But I'll eventually get one and investigate the issue.

Thanks for the feedback!

kkartaltepe:
>  I don't have a computer with an NVidia card at hand right now.

VAAPI would be using the intel chip (Since nvidia doesnt support open standards)

kkartaltepe:
> and the debian package for obs-websocket can simply be unpacked with a modified directory layout in $XDG_CONFIG_HOME/obs-studio/plugins and works as-is.

This is unsupported and seemingly unsupportable in both AppImage and Flatpak (as i understand it). So it seems more like a mark against both since users obviously want to be able to do something like use plugins safely (it just happened to not crash on your machine).

TiZ-HugLife:
> > and the debian package for obs-websocket can simply be unpacked with a modified directory layout in $XDG_CONFIG_HOME/obs-studio/plugins and works as-is.
> 
> This is unsupported and seemingly unsupportable in both AppImage and Flatpak (as i understand it). So it seems more like a mark against both since users obviously want to be able to do something like use plugins safely (it just happened to not crash on your machine).

At least for Flatpak, I'm pretty sure it's the contrary. Because Flatpak installations are kind of like overlay distros; when you install an app, you install the runtimes it depends on, and you have a second set of userspace libraries that Flatpak applications use. So if it didn't crash on my system, I'm confident it would not crash on any other system.

TiZ-HugLife:
Also, the maintainer for obs-websocket only distributing a debian package is pretty much [just on him](https://github.com/Palakis/obs-websocket/issues/442), since the linux browser plugin can distribute a distro-agnostic tarball just fine.

kkartaltepe:
> So if it didn't crash on my system, I'm confident it would not crash on any other system.

Thats only happenstance because the binary you are running was not built against the flatpak runtime and is not updated with the flatpak runtime. As such it could break at anytime if you happen to update it or the flatpak.

> Also, the maintainer for obs-websocket only distributing a debian package is pretty much just on him, since the linux browser plugin can distribute a distro-agnostic tarball just fine.

Im not sure what you mean by this. You could just unpack the deb and call it as agnostic the linux-browser tarball. Its just built against cef targeting ubuntu 14.04 and only supported on ubuntu https://bitbucket.org/chromiumembedded/cef/wiki/BranchesAndBuilding#markdown-header-legacy-release-branches-unsupported (Potentially more restrictions based on where the plugin itself was linked, maybe 16.04). If that was agnostic enough we wouldnt need appimage or flatpak.

TiZ-HugLife:
Well, ultimately, you are right. So as I'm sure you can see right above this comment, I've asked the maintainers for the current flatpak on flathub about plugin extension points. I'm glad I have obs-websockets working now, but I do want to use other plugins and we should figure out how to do it the right way.

GeorgesStavracas:
PR https://github.com/flathub/com.obsproject.Studio/pull/80 makes the JACK plugin successfuly work on top of PipeWire. You still need to have PipeWire 0.3 installed on your host system.

My personal expectation is that PipeWire 0.3 will be on most distros by next year (the biggest blocker was Debian, and I'm happy to say that the updated PipeWire package is getting close to land! See https://ftp-master.debian.org/new/pipewire_0.3.10-1.html and https://salsa.debian.org/utopia-team/pipewire/-/merge_requests/3). I also expect the ALSA plugin to go through the same route, eventually.

kkartaltepe:
The've got 6 or so months to get through the process, I hope they can do it.

GeorgesStavracas:
I'm happy to report that `pipewire` 0.3 is now part of Debian (currently, [Debian Experimental](https://tracker.debian.org/pkg/pipewire)) and Debian folks are enable PipeWire everywhere. Unfurtunately, I was let known that Ubuntu 20.10 won't add it, but from 21.04 forward it will.

PR #22, opened 2020-04-22
Add RFC: Generic outputs dialog

cg2121:
### Description
- Add generic outputs dialog so all outputs can be accessed at once, instead of being in the tools menu.
- This would be similar to the advanced audio dialog

Mockup:
Screenshot from 2020-04-21 22-57-21

### Motivation and Context
- With each output currently, developers have to create UIs for each one.
- These UIs may be hard to find in the tools menu
- This would simplify the output process and be more user friendly.

### Link to RFC
https://github.com/cg2121/rfcs/blob/output-dialog/accepted/0022-outputs-dialog.md

### DIFF
--- a/accepted/0022-outputs-dialog.md
+++ b/accepted/0022-outputs-dialog.md
@@ -1,0 +1,41 @@
+- Start Date: 2020-04-21
+- RFC PR: #22
+- Mantis Issue: N/A
+
+# Summary
+
+- Add generic outputs dialog so all outputs can be accessed at once, instead of being in the tools menu.
+- This would be similar to the advanced audio dialog
+
+# Motivation
+
+- With each output currently, developers have to create UIs for each one.
+- These UIs may be hard to find in the tools menu
+- This would simplify the output process and be more user friendly.
+
+# Detailed design
+
+## User UX
+
+- Display name of output in first column
+- Have a controls button to start and stop the output
+- An auto-start checkbox to start the output when OBS is started
+- Properties button to access the output properties
+	1. The properties dialog would be similar to the source properties dialog
+- Trash button to remove output
+- At the buttom have a plus button
+	1. The plus button would be used to add outputs, such as if the Blackmagic source allowed outputs for specific sources
+	2. The plus button would bring up a menu similar to the add source menu
+
+## How to teach this
+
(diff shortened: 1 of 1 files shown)


dodgepong:
> Add buttons for each output in controls menu. This would probably get cluttered fast.

I'm much more in favor of this idea, personally. Remove the Exit button, and move the settings and Studio mode buttons somewhere else, and turn that dock into a output controls dock.

I think it can be simplified quite a bit from the proposed UI, too. The Remove buttons can be omitted entirely, and I feel like the auto-start checkbox should be something inside the properties of the output rather than exposed front and center like that. Then the output label and start/stop toggle could be combined into a single field where the button itself says "Start <name of output>" and "Stop <name of output>". The more complicated bit would be handling the extra buttons like Pause on recording and the Save Replay button from replay buffer.

The other less clear thing in this UI proposal is what the cog wheel does when you push it. For Decklink and the other third-party plugins, it would presumably open the windows that are currently available in the Tools menu for those plugins. If outputs are to be combined into a more consolidated location, though, then we should take a look at revamping the Outputs tab under settings.

Unfortunately I think this would only alleviate one issue among a web of several inter-connected issues related to output management, and it's hard to see a way to not have to tackle them all at once.

cg2121:
This dialog, in this proposal, is for raw outputs, not encoded ones like streaming and recording. The stream, record and replay buttons would remain the same, and their settings would remain in the settings dialog.. The main reason is to take the raw outputs out of the tools menu, into a consolidated place.

I think the controls dock would get especially cluttered if there were outputs for individual sources, for example.

dodgepong:
I would argue that the Tools menu already is a consolidated place, but I don't think "encoded output" vs. "raw output" is a meaningful enough distinction to keep the two types separate. Furthermore, I don't see how separating the two types of outputs into completely different windows makes this less confusing than if they were all together.

As for clutter, this is a problem we're going to need to overcome anyway when we allow for multiple RTMP outputs, for instance. Suppose someone wants to stream to five different streaming destinations at the same time with OBS. How do we present those people with controls for all five of those outputs?

cg2121:
Closing because I think that this is not a good way to do this anymore. I think the best solution would be to allow plugins to add their own buttons to the controls dock.

PR #24, opened 2020-05-08
Add RFC: Improve the projectors and the multiview window

Bennik2000:
# Summary
Improve the usability and add more freedom for customization of projectors and the multiview window 

# Motivation
The projectors and the multiview windows are quite useful but do not provide much options for customization. 

# [Read RFC](https://github.com/Bennik2000/rfcs/blob/improve-multiview-and-projectors/accepted/0024-improve-multiview-and-projectors.md)

### DIFF
--- a/accepted/0024-improve-multiview-and-projectors.md
+++ b/accepted/0024-improve-multiview-and-projectors.md
@@ -1,0 +1,73 @@
+- Start Date: 2020-05-08
+- RFC PR: #24
+- Mantis Issue: N/A
+
+
+# Summary
+
+Improve the projectors and the multiview window.
+
+# Motivation
+
+The projectors an the multiview windows are quite useful but do not provide much options for customization. 
+
+# Detailed changes
+
+### Changes in the multiview window
+
+Add more freedom to customize the multiview:
+
+- Add some sort of configuration window which allows the user to set the following settings:
+  - number of rows 
+  - number of columns
+  - size of the preview in columns and rows
+  - show/hide the preview
+  - show/hide the program
+  - show/hide the name of the scene/source
+  - always on top
+- show/hide safe areas (EBU R 95) for the preview and scenes/sources
+
+- Add a right click menu to select which scene is visible in one region of the multiview
+- Add an option to transition to the scene by double clicking on one region
+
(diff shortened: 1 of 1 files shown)


univrsal:
When I was playing around with the projector code I had some additional ideas, so I'll just add them here.
Since the multiview is a grid, the user can right click on each cell of the grid and select what source or scene they want it to display (including the preview and program). There would also be an option to drag over multiple grid cells and then right click which would then display the source or scene spanning over the selected cells. This would make the multiview a lot more dynamic, but probably also a bit more complicated. By default the grid would probably be populated the same way it has been so it would be consistent with the current behavior.
I also thought it would be nice if the multiview could show more than just scenes and sources, for example audio levels or stream statistics, which would mean that a cell could also be populated with custom widgets, which handle the rendering.
These features might be a bit overkill but I thought I'd suggest them regardless.

unknowndomain:
I agree that the multiview could do with more configuration options and the idea of a grid is really nice because assuming your output is 16:9 and most monitors are 16:9 or 10 then having 16:9 ratio cells would make scaling and reorganising things really nice.

It would make sense to allow multiple multiviews so that you could have a three monitor setup, 1 for OBS, one for a grid of sources and one for a program/preview preview for example.

I3ordo:
Well i am looking forward to these, very needed improvements imo. You know the OBS layout is flexible, but have some shortcomings in very unexpected ways.
 for ex. 
Cant have fully vertical panels if you put any "dockable" under  or below the preview window
cant dock projectors
have borderless projectors.
Of course things seem too grim if you word it like this, in truth OBS is love!

Fenrirthviti:
Just for clarification, the limitations you listed are limitations of Qt, not OBS (save dockable projectors, we could probably do that).

univrsal:
Just wanted to give a little update, I've implemented some of the features for the multiview I mentioned in a [separate plugin](https://obsproject.com/forum/resources/durchblick.1484/).

Fenrirthviti:
After internal discussion, we feel that the additional functionality provided by the plugin is exactly the kind of thing people looking for that extra functionality have plugins available to do so.

PR #26, opened 2020-05-12
Add RFC: Add HLS output option to OBS

ushadow:
# Summary
  
Add HLS output as one of the live streaming options in OBS.

# Motivation

[HLS](https://tools.ietf.org/html/draft-pantos-hls-rfc8216bis-07) live streaming is supported by a number of platforms. It is a generic output method. FFmpeg has an HLS option as a muxer output: https://ffmpeg.org/ffmpeg-formats.html#hls-2. If a URL is specified, it will stream the output to the destination. Akamai has a [specification](https://learn.akamai.com/en-us/webhelp/media-services-live/media-services-live-encoder-compatibility-testing-and-qualification-guide-v4.0/GUID-6A14ED6D-0A23-4122-AB60-64A49B6628B5.html) for HLS ingestion as well. Many hardware encoders support HLS output. YouTube also supports HLS ingestion for live streaming and its [specifications](https://developers.google.com/youtube/v3/live/guides/hls-ingestion) are compatible with others.

HLS can support next generation codecs that RTMP does not. Newer codecs can offer much better compression relative to H.264, allowing users to stream with higher quality for a given bitrate, or stream with the same quality while using a lower bitrate decreasing buffering.

Currently one can do HLS output using OBS by setting FFmpeg HLS option in the recording output settings and starting to record. This is a bit convoluted though. So the goal of this RFC is to make the HLS output option more user friendly and more accessible.

### [Read RFC](https://github.com/ushadow/rfcs/blob/hls-ingestion/text/0026-hls-ingestion.md)


pkviet:
if the rfc is accepted, would you be the one who'd code it ?

pkviet:
and what would be the useful params not available in ffmpeg, which would make a native implementation the best course ?

dodgepong:
Is HLS ingest a generic output method, or is this proposal specifically for YouTube HLS? Is there a neutral spec for any company to follow that would enable them to accept HLS livestreams, or is the standard currently "do whatever YouTube does"? I can't tell if this RFC is proposing that OBS implement a general "HLS output" that would include YouTube, or just specifically implement "HLS output for YouTube".

It sounds like the primary motivation for this output method is to allow users to stream to YouTube using codecs other than h264, at the cost of much higher latency. Can you be more specific about which "next generation codecs" users would be able to use? We have no interest in promoting HEVC, for example. Are you referring to VP9, or is this in anticipation of wider availability of real-time AV1 encoders to consumers?

ushadow:
@pkviet If the RFC is accepted, we are planning to make it as an intern project at YouTube. YouTube engineers will review the changes before making a pull request. Upon closer inspection, the FFmpeg options for HLS seem to be pretty comprehensive. So at this point, I don't see a reason to make a native implementation.

ushadow:
@dodgepong HLS ingest is a generic output method. FFmpeg has an HLS option as a muxer output: https://ffmpeg.org/ffmpeg-formats.html#hls-2. If a URL is specified, it will stream the output to the destination.  Akamai has a [specification](https://learn.akamai.com/en-us/webhelp/media-services-live/media-services-live-encoder-compatibility-testing-and-qualification-guide-v4.0/GUID-6A14ED6D-0A23-4122-AB60-64A49B6628B5.html) for HLS ingestion as well. Many hardware encoders support HLS ingestion. YouTube's HLS ingestion requirements are compatible with all of them. In fact, currently one can do HLS ingestion to YouTube using OBS by setting FFmpeg HLS option in the recording output settings and starting to record. This is a bit convoluted though. So this RFC is to make the HLS ingestion option more user friendly.

Regarding codecs, for this RFC, we are just going to add support for HLS with H.264. This will lay the groundwork for adding support for additional codecs in the future. YouTube currently supports H.264 and HEVC as input codecs for HLS ingestion, but if there is user demand for VP9, we will add support for that as well, and the same is true for AV1 when it is ready. I'll add more clarifications in the RFC.

sperok:
Per the YouTube [HLS Ingestion documentation](https://developers.google.com/youtube/v3/live/guides/hls-ingestion) cdn.ingestioninfo is filled in after an authenticated call to LiveStreams.insert:

```
"cdn": {
    "format": "variable",
    "ingestionType": "hls",
    "ingestionInfo": {
      "streamName": "j7wz-wpku-wvu1-b2c4-3b6v",
      "ingestionAddress": "https://a.upload.youtube.com/http_upload_hls?cid=j7wz-wpku-wvu1-b2c4-3b6v&copy=0&file=",
      "backupIngestionAddress": "https://b.upload.youtube.com/http_upload_hls?cid=j7wz-wpku-wvu1-b2c4-3b6v&copy=1&file="
    },
```

What do you see as the UX for authentication and incorporation of the streamName and ingestionAddress?

Fenrirthviti:
Can you update the name of this PR to say "HLS output" instead of ingest? The way it currently reads, it sounds like you want to add the ability to input HLS in to OBS (which is already possible).

ushadow:
The ingestion addresses for HLS output will be set in the OBS's
services.json file, just like all the other services. Users will be able to
copy the Stream Key from YouTube’s creator Studio similar to YouTube RTMP
ingestion.

ushadow:
@Fenrirthviti That's a good point. I've updated the PR to say "HLS output".

derrod:
Regarding the documentation on YouTube's developer site, it says the following:

> The encoder must:
> * send exactly one encoded stream with the highest bitrate that you want to serve to users (single bitrate and codec).

This may be misinterpreted as YouTube actually offering a "Source"/pass-through quality option akin to Twitch, which appears to not be the case. Perhaps it's meant to say "resolution" instead?

ushadow:
@derrod Yes, we do mean resolution. We will update the documentation.

jp9000:
Seems reasonable to me, as long as congestion is handled (and it seems to be), then I'm okay with it. Later on, I might have to update ffmpeg-mux for more interprocess communication so it can also perform dynamic bitrate.

Going to put this in to final comment period, if anyone has any more comments please chime in.

Fenrirthviti:
FCP passed, merging.

PR #29, opened 2020-07-16
Add RFC: Controls API Implementation

cpyarger:
### Summary
Either Extend the obs-frontend-api, or create an obs-control-api

 ### Motivations
For plugins to control OBS, many of the same functions are currently required to be implemented individually by each plugin. My suggestion is to take many of the wrapper functions in use by OBS Websockets for interfacing with OBS and add them to OBS itself so they don't need to be reimplemented by the plugins.

This RFC may be used in conjunction with [RFC28](https://github.com/obsproject/rfcs/pull/28), but is completely UI independent.


cpyarger:
It has been suggested a couple of times I submit this as a separate RFC so here I am (: sorry about the extra commits above :/

cpyarger:
an example of what I am thinking is the GetSceneFromNameOrCurrent, and other "Utils" functions. 

also possibly the RequestHandler wrapper functions

ibex-are-goats:
Thank you for doing this important work!

DDRBoxman:
This RFC is currently a stub and needs more details before a discussion can happen.

cpyarger:
A lot of what I was looking for in this RFC has been added along with obs-websocket's integration into obs. So I am going to close this.

PR #31, opened 2020-07-20
Add RFC: Automatic Configuration Backup

derrod:
### Summary

Add built-in backup functionality to OBS Studio.

### Motivation

As of writing OBS does not offer any native option for users to have their settings (profiles and scene collections) automatically backed up to cloud, local, or network storage. At the moment backing up OBS' entire configuration data is also not easily possible or intuitively explained for non-technical users.

This RFC is meant to firstly facilitate a discussion to determine a high-level approach to implementing a backup feature in OBS Studio, before then formulating a specification for the low-level implementation of the chosen approach.

## [View the RFC](https://github.com/derrod/rfcs/blob/cloud-backup/text/0031-automatic-backups.md)

### DIFF
--- a/text/0031-automatic-backups.md
+++ b/text/0031-automatic-backups.md
@@ -1,0 +1,99 @@
+# Summary
+
+Add built-in backup functionality to OBS Studio.
+
+# Motivation
+
+As of writing OBS does not offer any native option for users to have their settings (profiles and scene collections)
+automatically backed up to cloud, local, or network storage. At the moment backing up OBS' entire configuration data
+is also not easily possible or intuitively explained for non-technical users.
+
+This RFC is meant to firstly facilitate a discussion to determine a high-level approach to implementing a backup
+feature in OBS Studio, before then formulating a specification for the low-level implementation of the chosen approach.
+
+# Detailed changes
+
+To start, I propose the following three approaches for discussion (though more should be suggested and discussed):
+
+ 1. Automatic export to user-specified storage location
+ 2. Integration of third party (cloud) storage providers
+ 3. Backup infrastructure and account system operated by the OBS Project
+
+Additionally, this RFC will lay out some ideas for how to integrate these backup/restore functionality into the OBS UI.
+
+## Overview of the three suggested variants
+
+### 1. Automatic export
+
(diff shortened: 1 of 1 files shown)


text/0031-automatic-backups.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Overview of the three suggested variants
+
+### 1. Automatic export
+
+Arguably the easiest approach is to add an option to automatically export/save OBS configuration files to an additional
+location when they are written. This could be configured by the user to be a synced folder (e.g. Dropbox or Google Drive)
+or a network drive, but could also be simply a different storage drive for redundancy.
+
+Versioned backups could be implemented by allowing backups to be sorted by date or time in the filename or folder path.
```
bj-mcduck, line 32: If it's a single file, just naming them with a datetime convention would do the trick.
If you do go down the path of having assets (optionally?) backed up as well, then would using Git be a better choice?

Not sure how many things would need to be included for each version as I don't know much about the settings files...

ifohancroft:
I feel like it's not OBS's job to Backup the configs and upload them to the cloud. However, if the XDG Basedir Specification is implemented fully, that would make things a lot easier.

I.e. At the moment, OBS just puts everything under $HOME/.config/obs-studio/ (whether it uses the home variable, or looks for the XDG_CONFIG_HOME env or just hardcodes the path, I don't know.

My point is everything is currently being put into the config folder, where the config, data, cache and runtime folders should all be utilized for the appropriate types of data.

CrAsHnBuRnXp:
Why can Streamlabs OBS do this so easily but OG OBS cannot?

Fenrirthviti:
This isn't really an appropriate place for this conversation, in the future, please use our Forums or Discord for questions like this.

To answer the question simply, Streamlabs is a service provider, we are a standalone application, and not a service. This complicates the backup process as we have more considerations to take in to account.

The longer answer also includes lack of development resources for anyone to work on this.

CrAsHnBuRnXp:
I apologize. Im not a frequent user of this website so im unfamiliar with how things work. When searching for a solution I cam across this and posted. Thanks for being kind enough to still provide the answer that I was looking for. Hopefully anyone like me that comes across this wondering the same will have their answer.

PR #32, opened 2020-09-10
Add RFC: Scene setup wizard

cg2121:
### Description
This proposal is for adding a scene wizard for new users when first loading OBS.

### Motivation and Context
Users often find it difficult to set up OBS.

### Link to RFC
https://github.com/cg2121/rfcs/blob/rfc-scene-wizard/accepted/0032-scene-wizard.md

### DIFF
--- a/accepted/0032-scene-wizard.md
+++ b/accepted/0032-scene-wizard.md
@@ -1,0 +1,36 @@
+- Start Date: 2020-09-10
+- RFC PR: #32
+
+# Summary
+
+This proposal is for adding a scene wizard for new users when first loading OBS.
+
+# Motivation
+
+Users often find it difficult to set up OBS.
+
+# Detailed Design
+
+The user will also be able to open the wizard in the scene collection menu.
+
+### Wizard page 1
+- Buttons that allow the user to select how they want to use OBS
+	1. "I just want to record my screen"
+	2. "I want to use a scene template"
+	3. "I will set up OBS myself" (if they select this then exit wizard)
+
+- The scene templates would be just pre-made scene collections
+
+### Wizard page 2
+- If the user selects scene template, show them different scene collections they could choose from
+- The scene collections will have image previews as well
+- If the user selects screen recording, ask them what monitor they want to record
+
+### Wizard page 3
+- Have the user select their microphone and webcam
+- The webcam would have a preview
+- The microphone would also show a volume meter
+
+# Additional Information
+
+- Should it be a seperate wizard than the auto-config? Or combined into a single wizard?


accepted/0032-scene-wizard.md
```diff
@@ -0,0 +1,36 @@
...
+	1. "I just want to record my screen"
+	2. "I want to use a scene template"
+	3. "I will set up OBS myself" (if they select this then exit wizard)
+
+- The scene templates would be just pre-made scene collections
+
+### Wizard page 2
+- If the user selects scene template, show them different scene collections they could choose from
+- The scene collections will have image previews as well
+- If the user selects screen recording, ask them what monitor they want to record
```
WizardCM, line 27: We'd have to be extra clear here for a couple reasons:
1. When people say "my screen", sometimes they mean "Photoshop" or a game.
2. On a laptop on Windows, "choosing a monitor" will result in no input, causing confusion (until the GPU is toggled)

accepted/0032-scene-wizard.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Wizard page 3
+- Have the user select their microphone and webcam
+- The webcam would have a preview
+- The microphone would also show a volume meter
```
WizardCM, line 32: I'd personally vote that this should be where we include other basic things like choosing default microphone filters (ideally, at least, a noise suppression filter & a compressor). We should also make it super clear to the user if their mic is peaking.

Speaking of peaking, I wonder if it'd be possible to determine what a device's system volume is in dB. Windows does have the UI for it, and being able to notify a user that their mic is "above the recommended range" (ie above 0dB). 
image

accepted/0032-scene-wizard.md
```diff
@@ -0,0 +1,36 @@
...
+
+# Additional Information
+
+- Should it be a seperate wizard than the auto-config? Or combined into a single wizard?
```
WizardCM, line 36: I think a combined wizard would be preferred, and steps should be clearly skippable. We don't want to overwhelm the user, and spamming them with new popups after each "mini-wizard" is complete would not be a good experience.

accepted/0032-scene-wizard.md
```diff
@@ -0,0 +1,36 @@
...
+# Motivation
+
+Users often find it difficult to set up OBS.
+
+# Detailed Design
+
+The user will also be able to open the wizard in the scene collection menu.
+
+### Wizard page 1
+- Buttons that allow the user to select how they want to use OBS
```
WizardCM, line 17: I think stylised buttons would be good - big icon, short text. Make it SUPER user-friendly - this'll be their first interaction with the program.

WizardCM (commented):
I've added some initial thoughts. I'll add more over time.

accepted/0032-scene-wizard.md
cg2121: One easy way to include the filters, is to have them in the scene template, which is just a scene collection.

cg2121:
One thing that would need to be implemented to make this work properly, is a containerized scene collection file. This way, image files, etc. can be added to the template and loaded seamlessly.

VodBox:
> One thing that would need to be implemented to make this work properly, is a containerized scene collection file. This way, image files, etc. can be added to the template and loaded seamlessly.

I have the beginnings of an RFC for this started.

PR #33, opened 2020-10-08
Add RFC: Replay buffer save frontend event

hgonomeg:
### Description
This  PR contains an RFC that suggests adding a new event to the frontend API

### Motivation and Context

Currently the API does not give a possibility to react to the event of the replay buffer being saved from a plugin.
It will allow for greater plugin functionality.

[See the RFC here](https://github.com/hgonomeg/rfcs/blob/master/text/0033-replay-buffer-save-frontend-event.md)


olek-osikowicz:
What a good idea. I wish I had this feature already.

jp9000:
For this RFC, there's not really anything to discuss so I'm just bypassing normal procedures and merging it.

PR #34, opened 2020-12-15
Add RFC: Program-neutral Plugin Manager

Xaymar:
### Description
Proposes a program-neutral plugin manager which can be created without requiring a new version of libOBS through the use of libOBS procedure handlers. This plugin manager does not address the user interface or URL integration portion, which are separate future RFCs.

### Motivation and Context
Users often fail to find or install third party plugins to libOBS, resulting in additional support workload for both libOBS maintainers and support, as well as the plugin maintainer and support. This reduces that factor by adding an integrated and hopefully failsafe way to provide plugins.

### Link to RFC
https://github.com/xaymar/obs-rfcs/blob/rfc1/text/0000-plugin-manager.md

### DIFF
--- a/text/0000-plugin-manager.md
+++ b/text/0000-plugin-manager.md
@@ -1,0 +1,461 @@
+# Summary
+Proposes a plugin manager for libOBS (OBS Studio, OBS.live, Streamlabs OBS, ...) that eases the installation and removal of plugins into libOBS.
+
+# Motivation
+At the current point in time with version 26.1.x, users still have to manually install plugins and search for these. This is usually prone to user errors, and often results in false bug reports and additional support work.
+
+# Design
+The Plugin Manager is a plugin to libOBS itself, requiring no changes to the libOBS API itself which may break ABI compatibility between versions. It uses the [procedure handler API](https://obsproject.com/docs/reference-libobs-callback.html#procedure-handlers) to provide other plugins or UI integrations a way to directly interface with the manager.
+
+## Plugin Manager API
+### Structure
+```
+struct obs_plugin_manager {
+	/** The API and Feature level that the plugin manager supports.
+	 *
+	 * Consists of two 32-bit unsigned integers stored with the current CPUs 
+	 * endianness, storing the API level in the bits [0...31] and the feature 
+	 * level in the bits [32...63].
+	 *
+	 * The API level increases only when incompatible API changes are made, such
+	 * as deleting or moving members. It should be considered unsafe to continue
+	 * using the API if a higher or lower API level is found.
+	 *
+	 * The feature level increases for additions to the current API level and
+	 * resets back to zero if a new API level is introduced. It is safe to use a
+	 * higher feature level than supported, but not a lower one.
+	 */
+	uint64_t api_feature_level;
+
+	// API Level 0, Feature Level 0
+
+// -- PROVIDER API -- 
+	/** Register a new plugin provider with the manager.
+	 * 
+	 * @param provider A pointer to a provider for plugins.
+	 * @param provider_size The size of the obs_plugin_provider_t structure.
+	 * @return true if successful, false if an error was encountered.
+	 */
+	bool (*register_provider)(obs_plugin_provider_t provider, size_t provider_size);
+
+	/** Count the number of providers currently registered.
+	 * 
+	 * @return The number of providers registered.
+	 */
+	size_t (*count_providers)();
+
+	/** Get the provider at the specified index.
+	 * 
+	 * @param idx 
+	 * @return A valid provider if the index is in range, otherwise 0.
+	 */
+	obs_plugin_provider_t (*get_provider)(size_t idx);
+
+	/** Get the provider described by the given id.
+	 * 
+	 * @param id A null-terminated string containing the identifier to find.
+	 * @return A valid provider if the id was found, otherwise 0.
+	 */
+	obs_plugin_provider_t (*find_provider)(const char* id);
+
+// -- DEPENCY API --
+	/** Register a non-plugin dependency.
+	 * 
+	 * This should be used to register things like libobs, qt, cef and other 
+	 * parts that are not a plugin into the dependency tree.
+	 *
+	 * @param id The dependency id.
+	 * @param version A C-string containing a version in the format "A.B.C.D", 
+	 *        with each component matching the regular expression /[0-9]+/.
+	 * @param location A UTF-8 C-string that contains the location for the given
+	 *        dependency. May be set to 0 to specify a location-less dependency.
+	 * @return true if the registration succeeded, false if there was an error.        
+	 */
+	bool (*register_dependency)(const char* id, const char* version, const char* location);
+
+	/** Check if a dependency can be fulfilled.
+	 *
+	 * Checks against an internal database (which also contains plugins) if a
+	 * dependency can be fulfilled with the currently installed dependencies and
+	 * plugins. 
+	 *
+	 * @param id The dependency id to try and find.
+	 * @param min_version The minimum version of the dependency to fulfill,
+	 *        failing if no version greater than it was found. This should be a
+	 *        C-string with the format "A.B.C.D" where each components matches
+	 *        the regular expression /[0-9]+/.
+	 * @param max_version Optional, specifies the upper limit for versions to 
+	 *        look for. Set to 0 if not needed.
+	 * @param location Optional, a pointer to a 'const char*' that will hold the
+	 *        location of the fulfilled dependency. Zeroed if it could not
+	 *        be fulfilled, or if the dependency has no location.
+	 * @return true if the dependency could be fulfilled, false if not.
+	 */
+	bool (*check_dependency)(const char* id, const char* min_version, const char* max_version, char** location);
+
+// -- PLUGIN API -- 
+	/** Register an installed plugin with the manager.
+	 * 
+	 * This does not perform any loading of the plugin until later, it only 
+	 * informs the plugin manager that a plugin is at a certain location. It
+	 * additionally registers the plugin into the dependency database for later
+	 * fulfillment of dependencies.
+	 * 
+	 * @param plugin An obs_plugin_information_t object that contains all necessary information about the plugin.
+	 * @return true if registering was successful, false if an error occured.
+	 */
+	bool (*register_plugin)(obs_plugin_information_t plugin);
+
+	/** Count the number of available plugins.
+	 * 
+	 * Only counts unique identifiers, not versions, locations or providers.
+	 * 
+	 * @return The number of registered plugins.
+	 */
+	size_t count_plugins();
+
+	/** Retrieve the plugin id at the given index.
+	 * 
+	 * @param idx The index to retrieve the identifier for.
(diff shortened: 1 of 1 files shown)


Xaymar:
There are several remaining problems to be solved:

* Who handles installation? Is it the provider, or the manager? Or both?
    * If the provider handles it, how do we know what plugins are installed, and where, and which versions?
* What about dependencies, who solves them? Provider, manager or both?
* Should we allow per-user plugin installations?
    * Risks privilege escalation from user space on _any_ platform if OBS is running as administrator.
* Should loading of plugins be performed with multiple threads?

VodBox:
So I really like the directory structure as proposed, and the plugin metadata format. Some form of that should absolutely be used in a package format for plugins, regardless of how the management or API is implemented in OBS.

Additionally, I like that this RFC is at least concerned with a local implementation as it's starting point, rather than what usually ends up being talked about (like repositories and discoverability of plugins in-app, but without actionable steps to actually producing an end product). While I think this doesn't tell the full story of implementing it, it has enough detail for someone to be able to start work, which is good.

My worries are a bit more fundamental to the ideas you've presented though.

We have no responsibility to alternative frontends, and I don't see what benefit we're supposed to get out of working on an implementation that is a frontend agnostic plugin.

Plugins shipping with additional Qt modules would ideally need to report their use of modules so that OBS can make decisions around whether it's compatible. That is entirely a UI concern. The only way that would work is if the API let a frontend tell it dependency versions to check against that only matter for that frontend. Having the management as part of the frontend would make it much simpler, since we can reliably make more assumptions about the environment.

Doing it as part of the frontend also results in less work overall, since we can include the logic or checks as part of the existing code path that initializes libobs. This also makes it much easier to do things like prompting the user before libobs loads if we've determined that a plugin is causing problems, like preventing loads, or repeatedly crashing within the last so many launches.

I am aware that doing these as frontend changes will mean we need make changes to libobs to support them (such as either setting a module as disabled to prevent it from loading, or providing an alternative obs_load_all_modules function with a callback to let the frontend determine if it wants to load the plugin), but I see no reason to limit ourselves in how we develop anything for the sake of other frontends when that only adds additional blockers and work for us.

The whole thing with providers and additional plugin structures seems to only exist to avoid making changes to the frontend or libobs, but that adds a whole lot of complexity that a frontend and libobs change would make much simpler. Why a provider system, when we could add code that runs for all plugins that reads a manifest associated with an installed plugin, and if that includes information about channels or where to find updates, we perform regular update checks (probably through some kind of cached reverse proxy for privacy reasons) and act on behalf of the plugin to update (probably after prompting the user for permission) or present options for switching channels. This also means all loading of plugins occur in the one code path, rather than being split between the frontend and a plugin, so maintenance is much easier.

The only benefits that doing things this way seems to present to me, is making sure other frontends get the same benefits of the work (which does not benefit us) and—since it doesn't depend on frontend/libobs changes—we could push this kind of plugin manager to the program separate from an update (which I do not believe is important to us).

Xaymar:
> [Third Party Frontend Support]

It is correct that OBS Project has no obligation to support Third-Party Front-Ends, however I saw no reason to modify an already implemented design just to spite others. There are also massive benefits to doing it this way, such as detaching the frontend of the plugin manager from the plugin manager itself, allowing it to be updated outside of the OBS Studio update intervals. Plugins would also be able to query for information without requring a link to obs-frontend-api.

> [snip] we need make changes to libobs to support them (such as either setting a module as disabled to prevent it from loading, or providing an alternative obs_load_all_modules function [snip]

Such a modification to libOBS is not needed as [obs_open_module](https://obsproject.com/docs/reference-modules.html#c.obs_open_module) already exists. This function is internally called by obs_load_all_modules, which also calls obs_find_modules.

> Why a provider system

See [other RFC about Steam distribution](https://github.com/obsproject/rfcs/pull/30). The provider system would allow us to support Steam Workshop for plugin distribution without much difficulty.

VodBox:
Yes, it is clear that in my cursory glance to remind myself of how the module loading code works that I confused what was internal private functions, and what were public. You are correct that it wouldn't require any libobs changes, and I was mistaken.

Regarding what you have said in your response to third-party frontend support, I've been thinking through this for a few days, and I don't see how we're supposed to win here.

From what you have said here, and based on what you've said in our Discord server, it seems clear that you've made this RFC submission after having already done at least some—if not most—of the work necessary in implementing the design as described.

The point of the RFC process is to discuss the merits of potential implementation details of an idea so that it can inform an actual implementation that gets made later. On an RFC, we should be questioning the design, any assumptions, or general approach to a problem and whether that is of benefit to the project. It is well within scope for us to naturally question things like whether frontend agnosticism is good for this particular kind of feature, and to weigh up the merits against alternatives that would be tied into the frontend.

Since you've already implemented and achieved the goals set out in the RFC, and have done so in a frontend agnostic way, you have given yourself the power—whether you intend to wield it or not—to shut down or undercut any meaningful conversation of such things by framing it as us simply looking to spite others.

If we determine that it'd be easier to maintain and achieve our goals with a simpler implementation that doesn't introduce new structures and paradigms, and that exists as part of our frontend, then you can say we're spiting the developers of other frontends, because it has now been framed as us rejecting an existing, working, frontend agnostic solution in favor of a brand new solution that would only work for us.

If we don't agree with the fundamental design or approach and we reject it because of that, then you could say we're spiting you because you've already put in the work, and we'd be rejecting not just the idea but the work you've done upfront.

If we request changes that would make it conform to our preferred design or approach, then you could say we're spiting you because we'd be asking you to redo work you've already completed.

You chose to implement a design with no communication with us. You opened up this RFC and titled the PR around being program-neutral, specifically setting up the discussion to be around whether we want a plugin manager which is program-neutral. When I brought up that discussion point, your primary justification was using that existing implementation to defend the design, and framing an alternative implementation proposal as spiteful.

This is not how we want to approach RFCs, and having already put in work should never be a reason cited for why something should be a certain way. We're here to discuss designs and ideas on their merits and benefit to the project. We should be looking at the benefits a frontend agnostic implementation could provide us—of which you provide only a couple, and I personally don't see as worth the extra code, maintenance and complexity—but your framing now makes further discussion difficult to handle in good faith.

I can only really come to the conclusion that with your entire approach to this RFC, you either fundamentally misunderstand the purpose of the process, or worse, you've intentionally submitted this RFC in bad faith.

Xaymar:
> [RFC created after Prototype]

You are correct that I have created this RFC after implementing a "working" prototype, and as you might have read on Discord that prototype also leaves a lot of questions yet to be answered. Excerpts from Discord (also publicly visible if you scroll up in #development):

image
image
image

> [RFC Process & Structure]

[I am well aware of how RFCs work for OBS Project.](https://github.com/obsproject/rfcs/blob/master/README.md) I am also aware that I submitted this as a draft and not ready for review, and I hope you are too. This RFC is to date incomplete and based on a barely working prototype, which leaves too many questions open.

> [Third-Party frontend support part 2]

As far as I am able to tell from what you've written, your arguments were:

1. ["We have no responsibility to alternative frontends, and I don't see what benefit we're supposed to get out of working on an implementation that is a frontend agnostic plugin."](https://github.com/obsproject/rfcs/pull/34#issuecomment-745734439)
2. It might be simpler to integrate this as part of the frontend itself (which I believe is RFC #4, making this RFC useless).

And my counter arguments were:

1. I have no reason to modify a design just to spite alternative frontends. However due to unclear wording, you might have misunderstood it as an argument against your second point as well - this was not my intention.
2. Plugin Manager, Providers and Frontend can be updated out of line from OBS Studio, requiring no major OBS Studio update.
3. Third-Party developers can provide their own plugin providers, enabling additional distribution markets.
4. It will be easier to support a wider variety of distribution methods, such as Steam.

> You chose to implement a design with no communication with us.

Partially correct. What I have done is implement something out of thin air, with no planning, design or RFC to back it up at the time. In fact the current "working" prototype leaves so many questions open that it might not even be considered a prototype - all it does is discover plugins in predefined directories, list them and their information, and load them on launch. And yes, I have reduced my contact with the OBS Project team after having been asked to reduce my interaction with it - so I've literally followed your own requests, and am being blamed for it.

> You opened up this RFC and titled the PR around being program-neutral, specifically setting up the discussion to be around whether we want a plugin manager which is program-neutral.

The reason for the title including the "Program-neutral" part is because #4 already discusses a plugin manager that is tied to the current frontend. It was included to explicitly set the RFCs apart by their nature, instead of having two nearly identical RFCs. Effectively this RFC only discusses the backend part, not the frontend part.

> [snip] you have given yourself the power—whether you intend to wield it or not—to shut down or undercut any meaningful conversation of such things by framing it as us simply looking to spite others.
> [snip] then you can say we're spiting the developers of other frontends [snip]
> [snip] then you could say we're spiting you because you've already put in the work [snip]
> [snip] then you could say we're spiting you because we'd be asking you to redo work you've already completed.
> [snip] you either fundamentally misunderstand the purpose of the process, or worse, you've intentionally submitted this RFC in bad faith.

Please refrain from making further unfounded claims. I have not submitted this RFC in bad faith, quite the opposite actually - otherwise this would have already been fully designed, implemented and released before even considering an RFC. As you can see, none of that has happened.

univrsal:
This seems like a step in the right direction to me. The plugin manager is a complicated component so splitting it into front-end and back-end would allow us to just focus on getting the latter right in this RFC.
As long as the back-end provides an adequate interface (which from what I can tell seems to be the case), the front-end can be worked out separately. Now about the problems mentioned
- Who handles the installation?
  - I'd say the provider sounds fair enough. How would the installation process go, though? Would the provider (or manager) get all the necessary information from ``obs_plugin_information``? Also would it be feasible to let the manager provide a default installation handler, but if the provider has a custom one, it'll call that instead?
- What about dependencies, who solves them? Provider, manager or both?
  - What kind of dependencies? libobs, obs-frontend-api and qt (at least the components used by obs) could be registered by the manager. What about additional dependencies though? On Windows I ship fftw3 for an audio visualizer. While it's pretty unlikely that multiple plugins will need fftw it should probably still be registered, right? How would different plugins depending on different versions work? And how would this work on other platforms? Lastly there's plugins like obs-ndi which require additional components to be installed separately, how would those work?
- Should we allow per-user plugin installations?
  - I'd say definitely. I don't quite get the privilege escalation issue, though. Isn't that already the case when obs is running as admin? On linux and mac I could see it being an issue when installing dependencies or a plugin globally, the former already has a way of asking for root rights when starting the virtual camera so I guess that could be something to reuse.
- Should loading of plugins be performed with multiple threads?
  - If it works, why not. Does libobs do anything that could cause issues when modules are loaded through multiple threads? Also on an unrelated note, can plugins be loaded/unloaded dynamically? Would be cool if you can enable a plugin and have the new source show up immediately.

I guess I provided more questions than answers, but maybe it'll be useful in some way. I'd really like to see this move forward. If we can get this to a point where an early prototype can be created we can probably also get some progress done on the other RFC. So thanks for writing this RFC up!

Xaymar:
I apologize for the lack of updates here, the past few weeks have been very busy. I will most likely split some parts of this into additional RFCs, as this one has grown to a size where it no longer covers just one thing, but many things.

Xaymar:
Split the plugin packaging format to #35.

Xaymar:
Closing as requested.

PR #35, opened 2021-02-01
Add RFC: Unified Plugin Packaging Format

Xaymar:
### Description
Proposes a unified structure for plugins to follow across all platforms, instead of the weird mixture OBS currently has across all supported platforms. Additionally some optional design choices are included, if these are necessary.

### Motivation and Context
Currently all platforms have their own way of shipping platforms, some even being different by the way they are installed (system-wide vs local-user), which results in higher difficulty when shipping cross-platform plugins. The current behavior also prevents some much needed future features from even working correctly.

### [Link to RFC](https://github.com/xaymar/obs-rfcs/blob/rfc2/text/0035-plugin-package-format.md)

### DIFF
--- a/text/0035-plugin-package-format.md
+++ b/text/0035-plugin-package-format.md
@@ -1,0 +1,174 @@
+# Summary
+A unified way for plugins to be packaged for all platforms.
+
+# Motivation
+Currently all platforms have their own way of shipping plugins, some even being different by the way they are installed (system-wide vs local-user), which results in higher difficulty when shipping cross-platform plugins. The current behavior also prevents some much needed future features from even working correctly.
+
+# Drawbacks
+Potentially more complicated than current logic.
+
+# Design
+All plugins should be shipped and loaded in the following format:
+
+- `/<name>/data`
+    The plugins custom data, such as localized text and images.
+- `/<name>/bin[/<obs version>]/<platform>/[.../]<name>[.<ext>]`
+    Binaries for a specific platform. Platforms may choose to do additional parsing of the path in order to support version selection, architectures, and bitness.
+- **or** `/<name>/bin/<name>[.<ext>]`
+    Binaries for an intelligent plugin manager which can intelligently pick out specific variants of the binaries to actually install.
+
+## Keywords
+### Keyword: `<name>`
+The unique name of the plugin which must match the actual module name, excluding the file extension. This name can be anything, as long as it is valid for the platform and/or container the plugin is in.
+
+### Optional Keyword: `<obs version>`
+Minimum OBS Studio version supported by the binaries in the directory in the format `A[.B[.C[.D]]]`, where each component is a positive integer number. If left out, assumes that all OBS versions are compatible.
+
+### Optional Keyword: `<ext>`
+A platform specific extension to denote dynamic libaries. A generic extension, such as `bin`, `obs` or similar may also be supported, depending on the final implementation, but this RFC does not specify that anything but the platform native extension for dynamic libraries. On platforms where no such extension is required, the dot prefixing the extension is also removed.
+
+### Keyword: `<platform>`
+The platform for which the binary is made for. Platforms may opt for additional parsing of the `[.../]` part in order to filter by platform version, architecture or bitness. By default the following platforms are available:
+
+#### `windows`
+The Windows platform supports the additional keywords `<version>`, `<arch>` and `<bits>`, which should be written as: 
+
+```
+/<name>/bin[/<obs version>]/windows[/<version>]/<arch>-<bits>/<name>[.<ext>]
+```
+
+- `<version>` is optional and defines the system version (i.e. 10.0.19042).
+- `<arch>` is required.
+- `<bits>` is required.
+
+This layout allows us to (optionally) specify a minimum platform version, a required architecture, and a required bitness. For example, with this we can filter for Windows 10 20H2 x86 64-bits with the following text: `/<name>/bin/windows/10.0.19042/x86-64/<name>[.<ext>]`, and as such prevent the plugin from loading on unsupported platforms automatically.
+
+#### `mac` or `macos`
+The Mac/MacOS platform supports the additional keywords `<version>`, `<arch>` and `<bits>`, which should be written as:
+
+```
+/<name>/bin[/<obs version>]/mac[/<version>]/<arch>-<bits>/<name>[.<ext>]
+```
+
+- `<version>` is optional and defines the system version (i.e. 10.13).
+- `<arch>` is required.
+- `<bits>` is required.
+
+This layout allows us to (optionally) specify a minimum platform version, an required architecture, and a required bitness. We can easily filter for specific Apple devices, and require certain OS versions for plugins, or even have M1 or X86 exclusive plugins.
+
+#### `linux`
+The Linux platform supports the additional keywords `<distro>`, `<version>` `<arch>` and `<bits>`, which should be written as:
+
+```
+/<name>/bin[/<obs version>]/linux[/<distro>[-<version>]]/<arch>-<bits>/<name>[.<ext>]
+```
+
+- `<distro>` is optional, and specific distributions may have a fallback to other values if no specific version is present.
+- `<version>` is optional, and describes the distro-specific version, for example:  
+    **On Ubuntu:** 18, 18.04, 19, 19.10, 20, 20.04, ...
+    **On Debian:** 9, 10, 11, ...  
+    **On FreeBSD:** 9, 9.0, 9.1, 10, ...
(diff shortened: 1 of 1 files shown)


Xaymar:
This may also affect #30 if OBS wants to have Workshop integration. Definitely affects #34 and #4 if accepted.

text/0000-plugin-package-format.md
```diff
@@ -0,0 +1,79 @@
+# Summary
+A unified way for plugins to be packaged for all platforms.
+
+# Motivation
+Currently all platforms have their own way of shipping platforms, some even being different by the way they are installed (system-wide vs local-user), which results in higher difficulty when shipping cross-platform plugins. The current behavior also prevents some much needed future features from even working correctly.
```
Xaymar: Typo: `platforms` -> `plugins`.

text/0000-plugin-package-format.md
```diff
@@ -0,0 +1,79 @@
...
+
+# Drawbacks
+None.
+
+# Design
+Plugins should be shipped and loaded in this format on all platforms from now on:
+
+- `/<name>/data`
+    The plugins custom data, such as localized text and images.
+- `/<name>/bin/<platform>/<arch>-<bits>/<name>.<ext>`
```
Xaymar: Consider making `<arch>-<bits>` optional, in order to support alternative ways to ship plugins (Lua, Python, multi-arch).

Xaymar:
I've rewritten some parts of the RFC in order to be much clearer on how the implementation should work. Additionally future additional "platforms", like "lua" (for Lua cross-platform plugins) can now also be supported with ease. I'd love to get some feedback on this.

text/0000-plugin-package-format.md
```diff
@@ -0,0 +1,111 @@
...
+### Keyword: `<bits>`
+Finally, the bitness of the system. This is an integer of any value, but the common numbers are:
+
+- `any`: Any bitness, see FatELF on Linux or Universal Binary on MacOS.
+- `32`: 32-bits
+- `64`: 64-bits
+
+Future bitness support is easily added by simply just increasing the number, so if we ever end up with 128-bits, just use 128.
+
+## Behavior when loading Plugins
```
WizardCM: I vote this to be extended a bit:

1. All built-in plugins should be loaded before anything else
2. Plugins in the new format should be loaded second, as they'll be in unique directories and therefore their structure/loading should be predictable
3. Plugins in the original/legacy format should be loaded last, and if any of them have a name matching a built-in or new-style plugin, should (by default) **not** be enabled (though a UI prompt somewhere to allow loading duplicate legacy plugins could be considered)

text/0000-plugin-package-format.md
```diff
@@ -0,0 +1,111 @@
...
+
+# Drawbacks
+Potentially more complicated than current logic.
+
+# Design
+All plugins should be shipped and loaded in the following format:
+
+- `/<name>/data`
+    The plugins custom data, such as localized text and images.
+- `/<name>/bin/<platform>/[.../]<name>[.<ext>]`
```
WizardCM: If I'm reading this correctly, only the file with the same name as the plugin's directory will be loaded directly, correct? In that case, say we're talking about obs-browser which includes a `libcef.dll`, that'd also be placed in `/<name>/bin/<platform>/[.../]` alongside `obs-browser.<ext>`, yes?

text/0000-plugin-package-format.md
```diff
@@ -0,0 +1,111 @@
...
+
+#### `linux`
+The Linux platform supports the additional keywords `<distro>`, `<version>` `<arch>` and `<bits>`, which should be written as:
+
+```
+/<name>/bin/linux[/<distro>[-<version>]]/<arch>-<bits>/<name>[.<ext>]
+```
+
+- `<distro>` is optional, and specific distributions may have a fallback to other values if no specific version is present.
+- `<version>` is optional.
```
WizardCM: I notice there's no elaboration on what "version" means on Linux. As every single distro will be different, maybe we should (as an example) include how it'd be done on Ubuntu, and when support for other distros is added, there is a basis to work off.

It's important to note that for Ubuntu currently, we [build a separate version](https://launchpad.net/~obsproject/+archive/ubuntu/obs-studio/+packages) of OBS for each release of Ubuntu as the OS ships with different versions of our dependencies.

WizardCM (changes requested):
I notice #34 covers min/max versions of OBS that individual plugins are compatible with, and that this RFC only mentions system-level versioning. If I'm reading this correctly, that means that a plugin can only be shipped with a single minimum version of OBS in mind, but multiple versions for multiple operating systems - say, one that's compatible with Windows 10.0.19041 (20H1) and 10.0.19042 (20H2) separately.

Personally, I think a plugin should be comfortable shipping a slightly older version of itself within the same ZIP for manual installation. We already know a lot of users won't immediately update to a major OBS version, and will instead stick to the previous version to ensure all bugs are found. Being able to download one ZIP that contains builds compatible with the last 2 minor OBS versions seems beneficial.

Obviously a built-in Updater in OBS would not care about such a structure, because it could just download a compatible ZIP rather than the latest, but I want to make sure our bases are covered.

text/0000-plugin-package-format.md
Xaymar: Yes, so for your example obs-browser, it will only attempt to load obs-browser.so and ignore all other files. Those files are left up to the plugin itself to load, if at all.

text/0000-plugin-package-format.md
Xaymar: That sounds better than the current text, I'll adjust it.

text/0000-plugin-package-format.md
Xaymar: Yes, I could not find a unified cross-distro way to identify what the platform version is to really define it. On Ubuntu and Debian, I found that the command `lsb_release -a` gives a reasonable result, though Debian for whatever reason returns text instead of version numbers. This needs input from people that actually know how to identify a platform version.

Xaymar:
> I notice #34 covers min/max versions of OBS that individual plugins are compatible with, and that this RFC only mentions system-level versioning. If I'm reading this correctly, that means that a plugin can only be shipped with a single minimum version of OBS in mind, but multiple versions for multiple operating systems - say, one that's compatible with Windows 10.0.19041 (20H1) and 10.0.19042 (20H2) separately.
> 
> Personally, I think a plugin should be comfortable shipping a slightly older version of itself within the same ZIP for manual installation. We already know a lot of users won't immediately update to a major OBS version, and will instead stick to the previous version to ensure all bugs are found. Being able to download one ZIP that contains builds compatible with the last 2 minor OBS versions seems beneficial.
> 
> Obviously a built-in Updater in OBS would not care about such a structure, because it could just download a compatible ZIP rather than the latest, but I want to make sure our bases are covered.

I couldn't really figure out a way to decide on a proper way to create something that works as one Zip while keeping all the restrictions in mind, so I went for a middle ground. I'm unsure how useful such a backport functionality for Platform versions even is, as I couldn't think of many applications of it. This might perhaps actually be better to be split off into its own RFC for plugin metadata, only leaving the platform, architecture, and bitness.

My initial idea here did not take into account that a future RFC may use split packaging, where data and binaries are in separate archives. If split packaging is used, the entire structure effectively collapses to only requiring a `/<name>/bin/<name>[.<ext>]` structure, as the binary package can be made unique for each platform.

text/0000-plugin-package-format.md
Xaymar: Seems like Debian is the only(?) outlier here with not giving a proper version and patch level.

Xaymar: I've added a better definition of the version keyword for linux now.

text/0000-plugin-package-format.md
Xaymar: Adjusted.

Xaymar:
I've clarified some regions now, including versioning across platforms. Additionally I added an ideal load structure for multi-arch platforms where universal/fat binaries are common.

Xaymar:
@WizardCM 
> Being able to download one ZIP that contains builds compatible with the last 2 minor OBS versions seems beneficial.

The updated RFC should now be able to do this, though it has the same "minimum only" restriction.

> Obviously a built-in Updater in OBS would not care about such a structure, because it could just download a compatible ZIP rather than the latest, but I want to make sure our bases are covered.

Also now taken care of. Managers/Updaters most likely have no reason to download binaries which do not match the Host OS, so that was a good point.

Xaymar:
If #41 is accepted, this RFC requires an update to the `<name>` keyword.

Xaymar:
Closing as requested.

PR #39, opened 2021-05-28
Add RFC: Service Overhaul

tytan652:
*I'm sorry to have stupidly pushed a PR about this suddenly without talking about it with you.
So now I tried to write this RFC, it's not perfect, but we are here to discuss about it.*

This RFC is now split into two RFCs. They are both quite dependent of each other.

### Summary
- Make OBS able to accept third-party service plugins
- Attribute a id for each services rather than a common one
- A service can be provided with multiple protocols
- Separate Twitch, Restream and YouTube integration and make them plugins

### Motivation and Context

With those changes, developer will be able to create third-party service plugins.
And also service integrations will be separated as plugins.

[Link to RFC](https://github.com/tytan652/rfcs/blob/service_overhaul/text/0039-service-overhaul.md)

[Link to an implementation of the RFC](https://github.com/tytan652/obs-studio/pull/15)


henke37:
Services kinda suck in OBS. They can't setup outputs directly. They can't offer extra features such as the RPC of rtmp. In fact, rtmp doesn't even have a formal password/key system! Each service *should* be able to do its own rtmp setup as desired.

I hope that this redesign helps combat this problem.

Another aspect that is missing is services being able to pick which server to use given certain conditions.

tytan652:
The plugin `obs-services` will mainly register services which uses stream key.

`obs-services2` is meant to be used to implement services with specific behavior, right now there is Dacast, and many one with  ingest.

Edit: Also there is also the ability to create and use third-party service plugin. Which can help for really specific case.
Edit2: They can sorta set up output with recommended settings stuff

henke37:
"Sorta set up" isn't good enough for me. I'd like the service to be an output factory, taking control from centralized code for maximum control over the output.

tytan652:
> They can't offer extra features such as the RPC of rtmp. In fact, rtmp doesn't even have a formal password/key system! Each service _should_ be able to do its own rtmp setup as desired.

- RPC falls under `obs-outputs` RTMP implementation, not services. Please don't mix up outputs with services.

- I don't know if `obs-outputs` or even stream services follow the [RTMP specification](https://www.adobe.com/devnet/rtmp.html). But if it's not in the spec, in my opinion there is no "obligation" to implement it.

- I don't know when the team decided to need  a "representative" from the stream services when it comes to make PR with purpose to add a services. But there was a period where they were not put by "representatives".

- But if they really needed to implement it a certain way they were free to push a PR or ask for an addition of a mechanism to add third-party plugins if they don't want to be stuck with OBS release cycle.

henke37:
Correct, RPC in RTMP would be a feature of the output plugin. But the service should be able of encapsulating it in the form of a nice usable api. Which includes providing the login.

tytan652:
They are already usename and password in the [service API](https://obsproject.com/docs/reference-services.html#c.obs_service_info.get_username), I think they are just not used except for Dacast and Custom services.
But I don't know if RPC is implemented.

henke37:
RPC is not implemented in the rtmp client that obs uses. At least not in a *good reusable way*. I'm honestly more concerned about websocket based apis and such. It'd be very nice if obs could do stuff like subscriber alerts without the aid of 3rd party websites.

tytan652:
> It'd be very nice if obs could do stuff like subscriber alerts without the aid of 3rd party websites.

Not implemented imperatively in OBS, but in a plugin meant to supply a complete integration of the stream service.
I think those type of things, could be one of the use case of the possibility to add a third-party service plugin by the user.

tytan652:
Update: This RFC draft is split into two RFCs. I put them in the same PR because they are dependent of each other.
If you want me to separate them in two PR tell me.

PS: It's also easier to work on both of them at the same time.

tytan652:
The RFC 38 is quite well proven with a try of implementation of the API and not the UI logic for now [here](https://github.com/tytan652/obs-studio/tree/protocol_api).

Right now the only visible benefit is services not shown if FTL or RTMPS are not built with OBS.

I don't make it a PR for because it needs discussion before making it.

tytan652:
Since https://github.com/obsproject/obs-studio/pull/5047 is merged I can really begin the phase 4 after making some rebase.

jp9000:
My apologies for the trouble there. Hopefully we can move things forward more easily now that it's merged.

tytan652:
Update about phases:
- Phase 1 has now his PR even if was not a good idea, next time I will talk about it to @jp9000 beforehand
  - https://github.com/obsproject/obs-studio/pull/5104
- Phase 2 is just missing the docs and waiting for RFCs and Phase 1 review
  - https://github.com/tytan652/obs-studio/compare/rfcs_phase_1...tytan652:rfcs_phase_2

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,196 @@
...
+
+It will also provide the ability for some stream services to be able to made their own plugin.
+
+*Since using the service API was not possible with the actual OBS, any breaking change in it (if there are) will not impact any plugin for OBS.*
+
+# Design
+
+## UI
+The Stream settings page will only contain a combo box listing registered services with a remade "Show All" option. So property view can be shown under it.
+The "short list" with "Show All" option, will be harcoded with a list of service id in the page code.
```
DDRBoxman: If a streaming service is added from a plugin it should always be on this "short list".

We'll also want to keep the behavior from the "common" field in the current services json.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,196 @@
+***This RFC is dependent on [RFC 38](0038-protocol-api.md).***
+
+# Summary
+
+- Make OBS able to accept third-party service plugins
+- Attribute a id for each services rather than a common one
```
DDRBoxman: How will id collisions be prevented?

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,196 @@
...
+- A service can be provided with multiple protocols
+- Separate Twitch, Restream and YouTube integration and make them plugins
```
DDRBoxman, line 6: 👍 making each of these a plugin is a big win in my opinion.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,196 @@
...
+We need to restore the usage of property views and the possibility to create and use service plugin.
+
+It will also provide the ability for some stream services to be able to made their own plugin.
+
+*Since using the service API was not possible with the actual OBS, any breaking change in it (if there are) will not impact any plugin for OBS.*
+
+# Design
+
+## UI
+The Stream settings page will only contain a combo box listing registered services with a remade "Show All" option. So property view can be shown under it.
```
DDRBoxman: What does "remade "Show All"" mean?

Maybe add some UI mockups.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,196 @@
...
+The "short list" with "Show All" option, will be harcoded with a list of service id in the page code.
+Any removed element will be reimplemented in a new way.
+
+(Same for auto-wizard)
```
DDRBoxman: Please elaborate here. Does the wizard need new APIs too?  eg. bandwidth testing, stream testing, etc.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,196 @@
...
+It will also provide the ability for some stream services to be able to made their own plugin.
+
+*Since using the service API was not possible with the actual OBS, any breaking change in it (if there are) will not impact any plugin for OBS.*
+
+# Design
+
+## UI
+The Stream settings page will only contain a combo box listing registered services with a remade "Show All" option. So property view can be shown under it.
+The "short list" with "Show All" option, will be harcoded with a list of service id in the page code.
+Any removed element will be reimplemented in a new way.
```
DDRBoxman: What elements are going away and how are they being reimplemented?

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,196 @@
...
+This plugin provide services with only one protocol, so it need to "register" service like YouTube multiple times (HLS & RTMPS).
+
+And also OBS is too heavily dependent of it.
+
+#### Service integrations
+Those integrations are added and use `rtmp-services` in a hacky way and not as plugins.
+
+### Idea
+#### `rtmp-services`
+This plugin will be put in a state of deprecation with a legacy behavior to not break immediately every setup. And be replaced and finally removed after one major release cycle.
```
DDRBoxman: From a UX perspective we should find a way to remove this plugin and have it be seamless to the user.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,196 @@
...
+
+Since OBS is heavily dependent of this plugin, many changes will be needed to remove this dependency.
+
+#### Service integrations
+
+Twitch, YouTube and Restream have their own integration in OBS with OAuth. But those need to be isolated as front-end plugin(s).
+
+With an updated front-end API for those and future third-party plugins.
```
DDRBoxman: What new APIs we need here should be enumerated.

They also probably don't need to become proper "frontend plugins" I imagine that since they do oath flow and display webpages we can figure out a way to pull that functionality into libobs in an abstract way that the UI can build on universally.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,196 @@
...
+
+#### `custom-service`
```
DDRBoxman: Is the idea here that the user can put in a uri like his://  srt:// or rtmp://?

How would this service handle authentication?

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,196 @@
...
+ 
+Those services will be able to provide multiple protocols so no more "Service - HLS" and "Service - FTL".
+
+If a certain protocol is not available, the plugin will not register the service if the service doesn't use another protocol.
+If it does, the missing protocol will not be shown. [Thanks to RFC 38](0038-protocol-api.md).
+
+The notion of maximum and recommended shall be distinguished.
+Maximum, supported resolutions, recommended settings will be ignorable with their respective checkboxes.
+
+Those services should have no specific behavior like ingest management.
```
DDRBoxman: We should make sure to account for a future change here where services can provide a URL that returns ingest URLs and streaming quality settings from a standardized protocol

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,196 @@
...
+
+But can list with which codec they are compatible if needed. [Thanks to RFC 38](0038-protocol-api.md)
+
+In the end, adding streaming service will only be a JSON addition for this plugin.
+Only improvements will be accepted in the code of this plugin.
+
+#### `obs-services2`
+This plugin is meant to provide a replacements for `rtmp_common` type for services who need a custom behavior.
```
DDRBoxman: If these share code like making a request to a url to get a server list we should add a way to do that from the json. 

Otherwise each of these services should each be split out into their own plugin.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,196 @@
...
+
+This plugin is meant to have implemented and register services which need a specific behavior like ingest with a unique id for each one.
+
+No JSON file will be used to create those services, except maybe for servers list.
+
+#### Note about the plugins
+- Like this if the service needs a specific behavior, you can 'transfer' it from `obs-services` to `obs-services2` and keep the same id. The change will be seamless for the user.
+- Services with integrations may have two services registered. The basic one and its counterpart with OAuth.
```
DDRBoxman: It might be cleaner to have a way to support both from the same service. This way users get confused less and there is less clutter.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,196 @@
...
+- Maximums are now separated from recommended settings and can be set for each protocol but no settings for any protocol. **And the protocol shall be set, there will be no fallback to RTMP**
+- Supported resolutions are separated to, it's usualy because of the player used by the service, so it's set for any protocol. **Need to add new matrix logic**
+- Recommended settings and can be set for each protocol but no settings for any protocol. **And the protocol shall be set, there will be no fallback to RTMP.** *Consider it as futureproofing for when new codec for other protocol like AV1 will come. RTMP spec tell that it only support H264.*
+
+### Service ID naming scheme
+- Only lower case letter
+- Hyphen `-` are only used to distinguish two service providing the same stream service but with noteworthy differences like NicoNico (free & premium) or Twitch and Twitch OAuth
+
+### VOD Track
+This option will only be shown if the combo service-protocol is compatible.
```
DDRBoxman: This probably needs additional work not specified here other than "just a checkbox"

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Things to be reworked
+- Verification with video and output settings to see (if not ignored) if the settings for the service are respected.
+
+# Drawbacks
+
+This will break old OAuth authentification.
```
DDRBoxman: We should find a way to not break existing authentication

text/0038-protocol-api.md
```diff
@@ -0,0 +1,92 @@
...
+Create a better management of outputs, encoders and services with a protocol centered approach.
+
+Byproduct: Add also the possibility to support third-party plugin which add protocol support with his own output.
+
+*Services addition is a part of [RFC 39](0039-service-overhaul.md).*
+
+# Design
+*This RFC does not consider that because a protocol support a encoder it should be implemented in the main project.*
+
+## Protocols
```
DDRBoxman: Generally making a list of protocols that then have a list of codes seems okay. 

I have some concerns related to how if a protocol is registered service / output plugins configure the protocols. I believe that API layer still needs to be defined.

For example the webrtc protocol will have extra api calls the the service plugin will need to make to drive the process to start the stream.

text/0038-protocol-api.md
```diff
@@ -0,0 +1,92 @@
...
+So if a service can't handle all the compatible codec, adding a list of supported codec to a service related to his protocol shall be considered.
+
+`rtmp-services` will be modified to obtain this behavior beforehand with only H.264, AAC and Opus as compatible codecs.
+
+## Outputs
+Adding to the API, the required¹ protocol attribute just a char with the protocol name which tell OBS with which protocol it can be used.
+Also remove the compatible codecs attributes because it is a part of protocol now.
+And output should be able to support any codec that the protocol.
+
+A output only support one protocol.
```
DDRBoxman: Does this limitation make sense?

Do you have to have a nvenc hls and an nvenc rtmp output now?

DDRBoxman:
I'll start off by saying that eve though there is some interdependency between these two RFCs they should be two separate PRs.

Currently both of these RFCs are lacking in details and seem to hand wave large changes that will end up being made alongside these changes.

I've put a number of comments above as a start.

text/0039-service-overhaul.md
tytan652: Id collision are prevented when registering and this is already in OBS Studio
https://github.com/obsproject/obs-studio/blob/master/libobs/obs-module.c#L790

tytan652:
Thank you for your reviews

> I'll start off by saying that eve though there is some interdependency between these two RFCs they should be two separate PRs.

Like I said in the first comment:
> If you want me to separate this info two PR tell me.

So just tell me explicitly and I will do it.

I will look your other reviews later today.

text/0038-protocol-api.md
tytan652: > Does this limitation make sense?

An encoder provide an encoder for only one codec, so I guess a stream output provide an output for only one protocol would make sense.

> Do you have to have a nvenc hls and an nvenc rtmp output now?

I think there is confusion between encoder and output objects.

I will try to add more precision to avoid confusion.

text/0039-service-overhaul.md
tytan652: The OAuth flow with CEF and the use of docks make me thought that they will need to be "frontend plugins". If it's not needed, then it will not be.

text/0039-service-overhaul.md
tytan652: >I wanted to make them separated to be able to extract those OAuth plugins from the PPA packages to provide them as a standalone package for Linux distro with unofficial packages.
>
>But this is not part of RFC and is completely off-topic. And can be done an other way.

We can make plugins providing both way in one service with a switch button in their properties views. And not providing the client ids will build them without the integration part.

text/0039-service-overhaul.md
tytan652: ~~Check this [addition](https://github.com/tytan652/rfcs/blob/service_overhaul/text/0039-service-overhaul.md#rtmp-common-to-new-id)~~

text/0039-service-overhaul.md
tytan652: I don't understand what you mean with the "VOD Track" and a "just a checkbox" ?

text/0038-protocol-api.md
tytan652: > For example the webrtc protocol will have extra api calls the the service plugin will need to make to drive the process to start the stream.

I think it shall be implemented by the output and service, the protocol is just `"specification" put as code`

RytoEX:
> > If you want me to separate this info two PR tell me.
> 
> So just tell me explicitly and I will do it.

Yes, please split these two RFCs into two PRs.  If one relies on the other, or they must both be considered or merged together, note that in the PR descriptions.

tytan652:
PR split, commit will be squash later and phases are moved elsewhere.

text/0038-protocol-api.md
tytan652: Many things have changed since.
- #45

text/0039-service-overhaul.md
tytan652: Will not be broken. I hope.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,462 @@
...
+
+This plugin will need a brand new services.json with new format, to register each streaming service with their own id. No more things like `rtmp-common` id.
+ 
+Those services will be able to provide multiple protocols so no more "Service - HLS" and "Service - FTL".
+
+If a certain protocol is not available, the plugin will not register the service if the service doesn't use another protocol.
+If it does, the missing protocol will not be shown. [Thanks to RFC 45](https://github.com/tytan652/rfcs/blob/protocol_api/text/0045-protocols-in-outputs-api.md).
+
+The notion of maximum and recommended shall be distinguished.
+Maximum, supported resolutions, recommended settings will be ignorable with their respective checkboxes.
```
WizardCM: The fewer checkboxes the better. Especially if the idea is to also have a properties view.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,462 @@
...
+Before in OBS, in the Stream settings page, there was a combobox with two choice:
+- Streaming Services
+- Custom Streaming Server
+
+Each one had his own property view because those two are the two actual registered official services.
+This behavior could enable third-party service plugin to exist.
+
+Nowadays in OBS, this page show the list of services with many new elements showed of hidden depending of the selection (like recommended settings). With also Twitch and Restream OAuth integrations. And no use of the property views provided by `rtmp-services`.
+
+We need to restore the usage of property views and the possibility to create and use service plugin.
```
WizardCM: While Properties view has its uses, I don't see much of a use case for it in services - there isn't much else a service *needs* displayed. The elements listed below are better suited in a more "curated" style of controls. Though I do agree they're currently very "all over the place" and that definitely needs to be improved. The service config JSON should provide all the relevant rules as to what elements are relevant.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,462 @@
...
+    - `"max audio bitrate"` (optional): maximum audio bitrate for this service.
+    - `"max fps"` (optional): maximum frame per second for this service.
+    - `"supported resolutions"` (optional): array of supported resolutions for this service.
+    - `"bitrate matrix"` (optional): array of maximum bitrate for a certain resolution/FPS combo for this service.
+      - `"res"`: resolution.
+      - `"fps"`: frame per seconds.
+      - `"max bitrate"`: maximum video bitrate.
+
+##### Issues with this format
+- The naming scheme is a mix of object name with space and some with underscores.
```
WizardCM: This has since been fixed.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,462 @@
...
+    "services": []
+}
+```
+`"services"` is an array of service objects.
+
+Service plugin like integrations one should have a JSON root that look like this:
+```json
+{
+    "format_version": 4,
+    "service": {}
```
WizardCM: This feels like unnecessary semantics, as an array can contain a single entry. For consistency and ease of use, I'd prefer just the plural.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,462 @@
...
+Those additions are added like extensions to the service object format.
+
+Getters for those customisations are implemented in the plugin itself.
+- In `obs-services`:
+  - `"id"` (required): service identifier, services are no longer identified by their name. Added because this plugin is multi-services.
+  - `"name"` (required): name of the service. Added because this plugin is multi-services.
+  - `"common"` (optional, false by default): this object is added in service objects of this plugin to identify which service in "common" or not. No third-party plugin should have this behavior.
+- In `obs-youtube`:
+  - Servers name are translated.
+  - `"name_suffix"` in `"servers"` (optional): suffix added to the server name after being translated.
```
WizardCM: What is a use case for such a suffix?

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,462 @@
...
+It will also provide the ability for some stream services to be able to made their own plugin.
+
+*Since using the service API was not possible with the actual OBS, any breaking change in it (if there are) will not impact any plugin for OBS.*
+
+# Design
+
+## UI
+The Stream settings page and the auto-wizard will only contain a combo box listing registered services with also the "Show All" option. And the property view of the service will be shown under it.
+
+A "common" service is a service that is shown in the default service list. To keep this attributes with the UI, a flag will be added to Service API to allow services to not be in this list, so by default third-party plugins will be shown without the need to click on the "Show all" option.
```
WizardCM: The idea here is precisely what it sounds like, that a service is "common" - that is, if a user is looking to stream using OBS, a "common" service is the most likely option they will choose.

This does however bring up a concern - we currently only differentiate between common / not common, curated by us. Outside of that, services are provided in a _very_ lengthy alphabetical list. Introducing the ability for users to install any variety of service plugins could grow this list significantly.

Especially with a plugin manager where a user could discover / install any number of service plugins, being able to easily sift through that list would be crucial. I'm unsure how this could be done though - categories? Groupings based on country? There are a lot of services out there (we provide 89 as it is). I wouldn't be against listing newly-installed services from plugins near the top of the list though.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,462 @@
...
+### Service integrations new flow
+
+#### How docks and CEF widget for OAuth will be created
+For the docks we could add the possibility to add them through the frontend API.
+
+Adding to the frontend API the possibility to generate the browser widget neutral on the system use (CEF). Also need to add a function asking OBS Studio is browser features are availlable.
+
+OR
+
+Maybe adding an API to `obs-browser` to allow plugins to create widget and docks it without using the frontend API as a bridge.
```
WizardCM: Browser dock capabilities as exposed to plugins would be a UI feature that is simply _enabled_ by the existence of obs-browser. From a plugin's/service's perspective, they shouldn't have to care whether obs-browser exists, just that browser capabilities are enabled.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,462 @@
...
+
+#### How docks will be added
+
+`OBS_FRONTEND_EVENT_FINISHED_LOADING` can be used when OBS startup to enable docks.
+
+Three new front-end event will be added:
+- `OBS_FRONTEND_EVENT_SERVICE_CHANGED` emitted when the service was changed (new id).
```
WizardCM: Specifically when the user switches between services, yes?

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,462 @@
...
+- `OBS_FRONTEND_EVENT_SERVICE_UPDATED` emitted when the service had a parameter updated (no id change).
```
WizardCM: Specifically, when a user has changed one of the available properties/settings for a plugin, yes?

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,462 @@
...
+}
+```
+`"service"` is the service object itself.
+
+##### Plugins specific additions
+Those additions are added like extensions to the service object format.
+
+Getters for those customisations are implemented in the plugin itself.
+- In `obs-services`:
+  - `"id"` (required): service identifier, services are no longer identified by their name. Added because this plugin is multi-services.
```
WizardCM: I am unsure how best to handle an ID - services sometimes change their name or address, and often no longer want it associated with the old name. This means the plugin author would have to keep that in mind and create an ID that identifies the service without.. naming it. Maybe this should be a UUID-type field that has to be generated?

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,462 @@
...
+- In `obs-youtube`:
+  - Servers name are translated.
+  - `"name_suffix"` in `"servers"` (optional): suffix added to the server name after being translated.
+
+Specific documentations about adding service in `obs-services` should be made.
+
+##### Service object format
+<sup>S</sup>: Should be required by any first-party plugin
+
+<sup>O</sup>: Optional in `obs-services`, can be required or not used at all in first-party one-service plugins depending on their implementations.
```
WizardCM: Due to the size & complexity of the available options here, it would be worth [JSON Schema](https://json-schema.org/) document.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,462 @@
...
+  - `"video"`<sup>O</sup>: video codecs supported by the service.
+    - `"any_protocol"`<sup>O</sup>: array of strings with codecs' name supported by the service for any protocol.
+    - *`"protocol name"`*<sup>O</sup>: (e.g. `"HLS":`) array of strings with codecs' name supported by only this protocol. This array is merged with `"any_protocol"`.
+  - `"audio"`: audio codecs supported by the service.
+    - `"any_protocol"`<sup>O</sup>: array of strings with codecs' name supported by the service for any protocol.
+    - *`"protocol name"`*<sup>O</sup>: (e.g. `"HLS":`) array of strings with codecs' name only supported by this protocol. This array is merged with `"any_protocol"`.
+- `"supported_resolutions"`<sup>O</sup>: array of strings with resolutions (e.g. `"1280x720"`) supported by the service for any protocol.
+  - `"with_framerates"`<sup>O</sup>: default to false. If false each entry shall be put as this "1920x1080", if true it will be put with a framerate as this "1920x1080@30". Only those framerate will be be considered supported.
+  - `"array"`(required): Array of resolutions
```
WizardCM: `"with_framerates"` should not be necessary. A basic regular expression (or even a contains check) would easily determine if a frame rate is included. This way `"supported_resolutions"` can just be an array, rather than an object with an `"array"` key. The simpler the JSON structure, the better.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,462 @@
...
+      - *`"codec name"`*<sup>O</sup> (requires to be put in `"supported_codecs"` firstly): (e.g. `"av1":`) bitrate for this codec. It overrides `"any_codec"`.
+  - `"audio_bitrate"`<sup>O</sup>: maximum audio bitrate. Can be set per codec.
+    - `"any_codec"`<sup>O</sup>: bitrate for any codecs.
+    - *`"codec name"`*<sup>O</sup> (requires to be put in `"supported_codecs"` firstly): (e.g. `"av1":`) bitrate for this codec. It overrides `"any_codec"`.
+- `"recommended"`<sup>O</sup>: recommended settings that become the defaults when using this service.
+  - `"keyint"`<sup>O</sup>: is now per video codec.
+    - `"any_codec"`<sup>O</sup>: keyframe interval for any codec.
+    - *`"codec name"`*<sup>O</sup> (requires to be put in `"supported_codecs"` firstly): (e.g. `"av1"`) keyframe interval for this codec. It overrides `"any_codec"`.
+  - `"bframes"`<sup>O</sup>: is now per video codec.
+    - `"any_codec"`<sup>O</sup>: b-frames for any codec.
```
WizardCM: I think `"all_codecs"` or `"*"` might be better?

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,462 @@
...
+`OBS_FRONTEND_EVENT_FINISHED_LOADING` can be used when OBS startup to enable docks.
+
+Three new front-end event will be added:
+- `OBS_FRONTEND_EVENT_SERVICE_CHANGED` emitted when the service was changed (new id).
+- `OBS_FRONTEND_EVENT_SERVICE_UPDATED` emitted when the service had a parameter updated (no id change).
+
+Each service integration plugin will react to at least the two first to add or remove his related docks.
+
+#### Manage Broadcast
+YouTube is not the only service that could have or need a "Manage Broadcast" button. So adding a way to make other service plugin able to use it.
```
WizardCM: This sort of thing would definitely need to be part of a first version of this. Plus handling the Start Streaming button to intercept it in certain situations (such as opening Manage Broadcast indirectly).

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,462 @@
...
+
+### Property API addition
+- Add a property to be able to show information with text label
+
+Since the service API is in a way unusable by third-party, any breakage will only impact OBS built-in service plugins ( so only `rtmp-services`).
+
+### Service integrations new flow
+
+#### How docks and CEF widget for OAuth will be created
+For the docks we could add the possibility to add them through the frontend API.
```
WizardCM: Additionally, it should be able to add a dock without making it visible - this is how most of the Twitch docks work, we add them but don't display them to avoid overwhelming the user with too many new docks at once. Any API for managing these docks should be able to decide if said docks are docked by default and where.

WizardCM (changes requested):
Additional to my comments, it's worth also noting some of the other things OAuth services do - Twitch is probably the main example here, it has a dropdown that allow a user to choose which Javascript plugins to inject. The OAuth plugin then injects that. I'm unsure if we should leave that up to the plugin developer to figure out, or if we should provide some sort of "when chat dock has loaded" event.

-----

I'll probably have more feedback soon, just wanted to get my initial thoughts written down. There's a lot to consider.

text/0039-service-overhaul.md
tytan652: Those names are used for YouTube servers whatever the protocol:
- "Primary YouTube ingest server"
- "Backup YouTube ingest server"

And for RTMP they add (legacy RTMP) at the end ("Primary YouTube ingest server (legacy RTMP)").

I wanted to add translations for those two names. And put (legacy RTMP) as a suffix to have 2 translation keys rather than 4.

I will add the use case on the RFC.

text/0039-service-overhaul.md
tytan652: I just made this based on [DDRBoxman comment](https://github.com/obsproject/rfcs/pull/39#discussion_r690007308).

text/0039-service-overhaul.md
tytan652: It was just a draft idea, I actually add more browser related method to the Front-end API.

I will remove it on the next push.

text/0039-service-overhaul.md
tytan652: I will remove it on the next push.

text/0039-service-overhaul.md
tytan652: Yes, I will add more precision.

text/0039-service-overhaul.md
tytan652: Yes, I will add more precision.

text/0039-service-overhaul.md
tytan652: Didn't know about this, I never used too heavily JSON. I will check this link.

text/0039-service-overhaul.md
tytan652: This symbol was completely out of my thoughts when I wrote this. And yes, it might be better.

text/0039-service-overhaul.md
tytan652: - How about maximum are limits that even a user can't bypass and recommendation are ignore-able.
- Or recommendation are used to set settings with the Auto Wizard and maximums are ignore-able.

text/0039-service-overhaul.md
tytan652: An UUID will maybe make the user unable to detect easily which service is used through scripting or even websocket. If one day user do want to detect it through plugins or scripts.

I added the use of ID to avoid the creation of `alt_name` from the actual JSON, UUID can also fix the issue but are not "user that may use scripting on it" friendly.

text/0039-service-overhaul.md
tytan652: Note: every service that are not common will have the flag added.

tytan652:
> Additional to my comments, it's worth also noting some of the other things OAuth services do - Twitch is probably the main example here, it has a dropdown that allow a user to choose which Javascript plugins to inject. The OAuth plugin then injects that. I'm unsure if we should leave that up to the plugin developer to figure out, or if we should provide some sort of "when chat dock has loaded" event.

`obs-twitch` will have the same feature implemented through a dropdown in the property view and `OBS_FRONTEND_EVENT_SERVICE_UPDATED` that will change the script.

I will soon add what I add in the front-end API in a more detailed way, based on the exiting PR about adding BrowserDock API.

text/0039-service-overhaul.md
dodgepong: The "common" flag has been a subject of considerable debate. Personally once there's a plugin manager, I would be much happier to distribute some default service plugins and then offer some sort of search for new services to "install" into OBS so that each user essentially curates their own "common" list. Personally I'd like to see the [monolithic services.json be split into individual components](https://github.com/obsproject/obs-studio/discussions/3858) which makes it extremely easy to distribute new service plugins that don't require custom oauth features.

text/0039-service-overhaul.md
tytan652: Will be removed thanks to JSON Schema conditional pattern matching that work exactly like needed to not need `"with_framerates"`.

text/0039-service-overhaul.md
tytan652: `"name_suffix"` in servers will may be removed and be replaced by a possibility to change the protocol name in the protocol drop-down (will not change the value). In YouTube case the drop-down will show "Legacy RTMP" rather than "RTMP".

text/0039-service-overhaul.md
tytan652: A one-way conversion is now in the RFC.

text/0038-protocol-api.md
tytan652: This RFC was moved to #45.

text/0039-service-overhaul.md
tytan652: Completely removed, and use 4 translations key.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,387 @@
...
+- Make OBS able to accept third-party service plugins
+- Register services with a unique id rather than a common one
+- A service can be provided with multiple protocols
+- Separate Twitch, Restream and YouTube integration and make them plugins
+
+# Motivation
+
+Actually even if OBS has the Service API, developer can't create third-party service plugin because there is no mechanism to use them at all.
+
+Before OBS in the Stream settings page, property views were used for services but with only two registered services `rtmp_common` which contain all services and `rtmp_custom` for custom servers.
```
pkviet: Before 'in' OBS

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,387 @@
...
+
+Nowadays in OBS, this page show the list of services with many new elements showed of hidden depending of the selection (like recommended settings). With also Twitch and Restream OAuth integrations. And no use of the property views provided by `rtmp-services`.
```
pkviet: 'Currently' in OBS ... showS ... with many new elements showN OR hidden in dependence ON the selection ...

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,387 @@
...
+
+This need to be refactored to re-introduce property views for the service and for the protocol output.
+
+This will also provide the ability for some stream services to be able to made their own plugin.
```
pkviet: Spelling: to be able to make

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,387 @@
...
+
+# Design
+
+## Settings UI
+
+Many elements of the settings UI related to services will be moved inside services properties views like:
+- Stream key field
+- Username and password fields
```
pkviet, line 26: For info, SRT will require a new field, encryption level (aes 128, aes 256 ...).
I'm preparing a PR for that. The SRT alliance plugfest showed me it was required by many vendors.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,387 @@
...
+- Text with clickable link (e.g., YouTube integration links)
+- Maximum and recommended settings information
+- "Get Stream Key" button
+- "More Info" button
+
+This streaming output will be selected through the settings windows and no longer auto-selected while starting the stream and its property view will be placed below the service properties view.
+
+The service output settings (if any) will be saved in the `service.json` file.
+
+Advanced network settings will also be replace by the output properties view below the service properties view, since those are meant for `"rtmp_output"` (RTMP(S) output).
```
pkviet: Typo, replaceD

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,387 @@
...
+Maximums and supported resolutions will be strict (e.g. Twitch allows 6000 kbps, more will not be accepted), the UI will not allow to save incompatible settings.
+
+This will require to introduce in the Property API a way to indicate that the settings are not valid.
+
+TODO: Think about recommendation
+
+## Frontend API
+
+Those functions will be modified:
+- `obs_frontend_set_streaming_service()` because the output selection is done in the settings windows and is no longer done while the stream is starting will now swap the output if the service protocol and the actual output does not match.
```
pkviet: because ... while the stream is starting, this function will now replace the output to a compatible one if the service protocol and the output DO not match.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,387 @@
...
+Services provided by this plugin (`"rtmp_custom"`, `"rtmp_common"`) will be deprecated (`OBS_SERVICE_DEPRECATED` and `OBS_SERVICE_INTERNAL` applied) and completely unused in OBS Studio.
+
+Those are deprecated rather than completely removed to allow scripting and plugins to migrate if they happen to use them.
+
+Its service JSON will no longer be updated.
+
+
+### `custom-services`
+
+If OBS Studio need to be heavily dependent to one service plugin, it must be this one.
```
pkviet: If ... needS to heavily depend ON a single service plugin, it should be this one.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,387 @@
...
+
+This plugin is meant to provide replacements for the `"rtmp_custom"` type.
+
+#### `"custom_service"`
+
+This the only service of this plugin thar will be exposed to the user through the UI.
+
+The protocol will be selected by the user and the properties view will change depending of the protocol.
```
pkviet, line 122: will change according to the protocol.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,387 @@
...
+
+This service will all first-party protocol, the properties view will change visible fields depending on the protocol.
```
pkviet: Missing verb.
...will hold (?) all first-party protocolS.
The properties ... fields according to the selected protocol.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,387 @@
...
+
+And also support any third-party protocol by allowing to choose which fields (e.g., username + password, stream key, custom field…) to add in the view.
```
pkviet: ... protocolS

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,387 @@
...
+
+This service will be registered in the "post load module" step (`obs_post_load_modules()`) to make the service aware of all registered protocol in the "load module" step.
```
pkviet: ... protocolS

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,387 @@
...
+#### Custom type per protocol
+
+Service types that support only the protocol inside their id (`"custom_`*insert lowercase protocol acronym*`"`) for any first-party protocol except FTL (deprecation).
+
+Services that will only support the first-party protocol related to their id/type.
+The flag `OBS_SERVICE_INTERNAL` will be applied to not show them in the UI list. Those are meant for being used in WebSocket and scripting.
+
+### `obs-services`
+
+This plugin is meant to provide a replacements for `"rtmp_common"` type for services who doesn't have custom behavior or integration.
```
pkviet: ... a replacement (no S) ... for services which don't rely on an integration nor on customizations.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,387 @@
...
+**TODO: Some user might have a removed service (e.g., Mixer)**
+
+**Downgrade might break service integration**
+Auth integration config inside `basic.ini` will be also transfered under a new section if in service settings.
+
+### Integrations
+
+Twitch, YouTube and Restream have their own integration in OBS with OAuth. But those need to be isolated as plugins.
+
+Those will plugins will provide the basic and OAuth version of the services as one services.
```
pkviet: These plugins will ... (Will repeated twice)

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,387 @@
...
+
+If OBS is built without client ids and hashes, those plugins will be built without their integration.
+
+#### Dock state
+
+Actually the dock state is global in the config, but with docks that appear depending on the service which is per profile which is still "also" the case with "old" integrations.
```
pkviet: I had a hard time parsing this sentence ...
Split it please:
... which is per profile. This is also the case ...

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,387 @@
...
+
+This can lead to dock state losses between profile switching and exiting. The "old" integration actually store a dock state in the profile config and restore it after the integration is is loaded.
```
pkviet: ... is is loaded => is loaded

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,387 @@
...
+
+But making restoring a dock state from a plugin should should not be considered at all.
```
pkviet: But restoring a dock state from a plugin should => remove 'making' and extra 'should'

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,387 @@
...
+The plugin will have to store if `OBS_FRONTEND_EVENT_FINISHED_LOADING` was passed or not.
+
+If created before `OBS_FRONTEND_EVENT_FINISHED_LOADING`, a front-event callback will be added to react to to later event to add docks (if integration connected).
+
+When created/updated (after `OBS_FRONTEND_EVENT_FINISHED_LOADING`) the service will add the docks (if integration connected).
+
+When `OBS_FRONTEND_EVENT_PROFILE_CHANGING` `OBS_FRONTEND_EVENT_EXIT` or the service is destroyed, docks will be removed (if integration connected).
+
+#### Browser features
+The Front-end API needs to enable the possibility to access some `obs-browser` related feature like adding browser docks and generating widgets.
```
pkviet: Add some words elaborating the purpose.
What's the purpose of that feature ?
Do services really need it ?
(Auth, stream management, chat)

pkviet (commented):
Some spelling/typo corrections to easen the reading.
I'll come back later to a more general in depth review.

text/0039-service-overhaul.md
tytan652: `must` is the right wording.

text/0039-service-overhaul.md
tytan652: The `behavior` word needs to stay or have a good replacement, `customizations` is not enough.

text/0039-service-overhaul.md
tytan652: > Do services really need it ?

Except if you want to drop Twitch YouTube and Restream docks or create non-browser based alternative of those docks.
Otherwise yes those feature are needed since we completely stop having integrations directly inside the UI but as plugins.

I will elaborate the text in the future.

tytan652: > create non-browser based alternative of those docks

For docks like chats, this would take a while to implement.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,435 @@
...
+Any services that only support FTL will follow the deprecation cycle and be dropped at the same time as FTL if those services did not added support for another protocol.
+
+## Service plugins
+### Service API
+
+Adding to `obs_service_info`:
+  - `uint32_t flags` with the following flags:
+    - `OBS_SERVICE_DEPRECATED`: The service is marked as deprecated.
+    - `OBS_SERVICE_INTERNAL`: The service is meant to be used internally in some plugin (e.g., WebSocket, Scripting) and usually not directly exposed in the UI.
+    - `OBS_SERVICE_UNCOMMON`: The service can be hidden behind a "Show All/More" option UI/UX-wise.
```
DDRBoxman, line 66: Does it make sense for this to be up to the plugin. Maybe OBS should keep a list of common plugin IDs.

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,435 @@
...
+
+Any services that only support FTL will follow the deprecation cycle and be dropped at the same time as FTL if those services did not added support for another protocol.
+
+## Service plugins
+### Service API
+
+Adding to `obs_service_info`:
+  - `uint32_t flags` with the following flags:
+    - `OBS_SERVICE_DEPRECATED`: The service is marked as deprecated.
+    - `OBS_SERVICE_INTERNAL`: The service is meant to be used internally in some plugin (e.g., WebSocket, Scripting) and usually not directly exposed in the UI.
```
DDRBoxman, line 65: Can you expand on this with a more concrete example? I'm not clear on how web socket and scripting plugins quite match up to a service?

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,435 @@
...
+  - `bool (*can_bandwidth_test)(void *data)`: Return if the service is able to do bandwith test
+  - `void (*enable_bandwidth_test)(void *data, bool enabled)`: Enable bandwidth test on the service
+  - `bool (*bandwidth_test_enabled)(void *data)`: Return if the service has the bandwidth test enabled
+  - `void (*get_supported_resolutions2)(void *data, struct obs_service_resolution **resolutions,size_t *count, bool *with_fps)`: Replace its non-two variant to enable framerate value
+  - `int (*get_max_video_bitrate)(void *data, const char *codec struct obs_service_resolution resolution)`: Return a maximum bitrate based on a video codec and a resolution
+  - `int (*get_max_codec_bitrate)(void *data, const char *codec)`: Return a maximum bitrate for a specific codec
+  - `void (*apply_encoder_settings2)(void *data, const char *encoder_id, obs_data_t *encoder_settings)`: Replace its non-two variant to enable settings per encoder id and codec
+
+Adding to the Services API:
+  - `enum obs_service_audio_track_cap obs_service_get_audio_track_cap(const obs_service_t *service)`: Return the service audio track capability
```
DDRBoxman: Probably worth spelling out all the way

text/0039-json-schemas/service.json
```diff
@@ -0,0 +1,255 @@
...
+            "description": "Array of server objects",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "name": {
+                        "type": "string",
+                        "description": "Name of the server (e.g. location, primary/backup), displayed in the Server dropdown.",
+                        "minLength": 1
+                    },
+                    "protocol": {
```
DDRBoxman, line 32: Just a thought that different protocols might want different encoder settings

text/0039-service-overhaul.md
```diff
@@ -0,0 +1,435 @@
...
+The protocol will be selected by the user and the properties view will change depending of the protocol.
+
+This service will support all first-party protocols, the properties view will change visible fields according to the protocol.
+
+#### Custom type per protocol
+
+Service types that support only the protocol inside their id (`"custom_`*insert lowercase protocol acronym*`"`) for any first-party protocol except FTL (deprecation).
+
+Services that will only support the first-party protocol related to their id/type.
+The flag `OBS_SERVICE_INTERNAL` will be applied to not show them in the UI list. Those are meant for being used in WebSocket and scripting.
```
DDRBoxman, line 131: What does the migration path look like for scripts / apps using the old method?

Can we just provide a built in way to identify the old behavior and translate to the new one instead of keeping these around?

DDRBoxman (commented):
I think figuring out how to split this up into smaller pieces that we can incrementally build on instead of a giant overhaul would be a good thing to look into.

Especially if we can make the underlying parts divorced from the UI changes.

As a generality OBS could benefit from getting away tight UI coupling as we build features into OBS rather than building based on the UI we have today.

text/0039-service-overhaul.md
tytan652: `rtmp_common` and `rtmp_custom` are not removed but deprecated.
At some point we have to remove them.

text/0039-service-overhaul.md
tytan652: Third-party services are also common by default, and except if we want to add more hardcoded/duplicated list in the UI for uncommon services this is the only way I seen.

text/0039-json-schemas/service.json
tytan652: I try to avoid making a mess, and protocol-agnostic per codec settings is the simplest and the less messy.

Edit: If I could avoid doing `video_bitrate_matrix`, I would avoid it.

tytan652: If the service wants to do more than what `obs-services` allows, it can create a third-party service plugin.

tytan652:
> I think figuring out how to split this up into smaller pieces that we can incrementally build on instead of a giant overhaul would be a good thing to look into.

We can't since service integration is too tied to the UI and the UI is too tied to `rtmp-services`.

Edit: Once you touch a piece, you need to replace a whole block.

DDRBoxman:
It's probably worth doing an investigation to see if there are parts of the UI we can decouple as a prerequisite before the service overhaul. Smaller changes help lower the risk and allow easier review.

tytan652:
> It's probably worth doing an investigation to see if there are parts of the UI we can decouple as a prerequisite before the service overhaul.

Beside the YouTube Actions UI form, most of the service UI is refactored from Qt to PropertiesViews.  So there is not much to decouple.

text/0039-service-overhaul.md
tytan652: Changed the spelling a little to explicitly be about `rtmp-services` actual state, but the "future change" is out of scope of the RFC

text/0039-service-overhaul.md
tytan652: This separation was removed, it still now a unique checkbox in the OBS Studio UI.

text/0039-service-overhaul.md
tytan652: Completely overhauled, for a feature flag `OBS_SERVICE_UNCOMMON` in the Service API.

text/0039-service-overhaul.md
tytan652: Feel free to open a new conversation about this if needed.

text/0039-service-overhaul.md
tytan652: URI scheme detection was dropped in RFC 45 and this RFC will not add such detection. The user will have to explicitly set the protocol in use.

text/0039-service-overhaul.md
tytan652: The service got renamed to `orphaned-services` so not split because maintenance by us is not guaranteed.

text/0039-service-overhaul.md
tytan652: Audio track capability was added to the Services API.

text/0039-service-overhaul.md
tytan652: Using PropertiesView is required to isolate services as plugins and enable third-party plugins.

tytan652:
I made a cleanup around old review conversations, some of them were closed with a comment some not because the RFC has really changed since then.

Fenrirthviti:
Based on feedback on this thread, and in off-thread discussions, the project maintainers agree that this RFC needs to be scaled back and broken up in to smaller parts to approach in a more incremental way.

The project largely agrees with the end goal of what this RFC is proposing: a more customizable, flexible, and pluggable state for services and outputs in OBS.

However, the size and scope of the current approach makes these changes difficult, if not near impossible, to do a complete review and provide specific feedback. It is so large in scope that it is unapproachable to others seeking to review it, and is not likely to be accepted in the current state.

It was not an easy decision, but if we want to start making progress towards this goal, a new approach is necessary. 

To that end, a better approach would be to split out the changes in to separate RFCs/PRs. We suggest a split along the following topics:

  - Start by taking the YouTube, Twitch, and Restream integrations and make them pluggable. This would be an invisible change to the user.
  - Once complete, evaluate where the overlap is between those plugins, and migrate functions in the OBS API to reduce duplication. Again, invisible to the user, no UI changes.
  - Once parity has been reached with current implementation, start work towards refactoring a new UI and adding functionality as distinct and separate tasks.
  
We again thank everyone for their work on this RFC, and we hope that with these changes we can reach some of its goals sooner rather than later.

EwoutH:
Hi! I'm curious, has there been any progress on this feature? Is there maybe a tracking issue I can follow?

I would like to separately record my screen and webcam with shared audio, which would benefit from the modular architecture being discussed.

Fenrirthviti:
This RFC is closed, and there has not been anyone taking up work as outlined in my last comment above. We welcome anyone who is willing to follow the general approach as described.

tytan652:
I'm slowly making a rewrite, https://github.com/tytan652/rfcs/tree/service_overhaul_neo

And really mean slowly, the addition of TEB upstream added a layer of complexity to take in account and I want to avoid burning myself out so I really take my time.

PR #40, opened 2021-06-03
Provide more nuanced control for when media sources start playing

cbeer:
# Summary

Provide more nuanced control for when media sources start playing

# Motivation

FFMpeg media sources currently either autoplay as soon as the OBS session is started, or (if the media source is configured with "Restart playback when source becomes active") will start/restart when the scene is activated.

VLC media sources have slightly more nuance, with visibility behaviors for (re)start/stop, play/pause, or always play. 

This RFC proposes to expand on the nuance from the VLC media source to:
- provide the user better control of the media behaviors
- provide consistency between the media playing sources

This nuance supports a few unmet use cases, including:
- fully manual playback of media sources 
- hybrid controls (automatic starting/manual stop, or manual start / automatic pause)
- "pause when not visible/unpause when visible"-parity for FFMpeg sources.

## [Link to the RFC](https://github.com/obsproject/rfcs/blob/9ce0b82c7287eff760e8e433e3ac872b35cca631/text/0040-media-show-hide.md)

### DIFF
--- a/text/0040-media-show-hide.md
+++ b/text/0040-media-show-hide.md
@@ -1,0 +1,44 @@
+# Summary
+
+Provide more nuanced control for when media sources start playing
+
+# Motivation
+
+FFMpeg media sources currently either autoplay as soon as the OBS session is started, or (if the media source is configured with "Restart playback when source becomes active") will start/restart when the scene is activated.
+
+VLC media sources have slightly more nuance, with visibility behaviors for (re)start/stop, play/pause, or always play. 
+
+This RFC proposes to expand on the nuance from the VLC media source to:
+- provide the user better control of the media behaviors
+- provide consistency between the media playing sources
+
+This nuance supports a few unmet use cases, including:
+- fully manual playback of media sources 
+- hybrid controls (automatic starting/manual stop, or manual start / automatic pause)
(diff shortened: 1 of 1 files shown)


eric:
I would also like to see the first frame be visible in Preview for local content and the live view available for remote content.

It would also be nice to have the playback controls work on media when Previewed.

Fenrirthviti:
So, there are a few parts of the RFC that seem fine in concept, and we actually have open PRs for (https://github.com/obsproject/obs-studio/pull/3269, https://github.com/obsproject/obs-studio/pull/6112, and https://github.com/obsproject/obs-studio/pull/6239 are examples).

However, this RFC as currently written as both too high-level concept, and too specific to a desired workflow, with a lot of gaps and this feels to me like it's more of an [Ideas](https://ideas.obsproject.com/) than an actionable RFC unfortunately.

Personally, I dislike the VLC source and don't like that it exists, as it was more of a stopgap solution to adding features for media playback since nobody had the time to do a refactor/rewrite of the media source itself. I would prefer we ditched the VLC source and fixed the problems in media source that lead to us using VLC in the first place. The above PRs move us closer to that. This also touches on some of the issues we have with sources themselves (i.e. readiness states and show/hide callbacks, etc.).

Given the above, I'm going to start FCP with disposition close.

PR #41, opened 2021-06-08
Add RFC: Reverse Domain Name Notation naming convention

Xaymar:
### Description
Adopt the [Reverse Domain Name notation](https://en.wikipedia.org/wiki/Reverse_domain_name_notation) naming convention to reduce potential collisions.

### Motivation and Context
Naming for additions has no guarantee that future additions to OBS or plugins do not eventually collide and corrupt setups. A unified naming convention that by design already prevents this should be adopted and enforced.

### [Link to RFC](https://github.com/xaymar/obs-rfcs/blob/rfc3/text/0041-reverse-domain-name-notation.md)

### DIFF
--- a/text/0041-reverse-domain-name-notation.md
+++ b/text/0041-reverse-domain-name-notation.md
@@ -1,0 +1,96 @@
+# Summary
+Adopt the [Reverse Domain Name notation](https://en.wikipedia.org/wiki/Reverse_domain_name_notation) naming convention to reduce potential collisions.
+
+# Motivation
+Unified naming conventions reduce the chances of name collisions in first-/second- and third-party additions/changes.
+
+# Drawbacks
+- This may break existing setups unless a transition layer is implemented to transfer old to new identifiers.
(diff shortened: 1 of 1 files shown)


Xaymar:
Partial change extracted from #34 and #35 in order to reduce massive RFCs into smaller ones.

WizardCM:
I personally like this sound of this change, but I feel a migration path / backwards compatibility would need to be fleshed out extensively. For example, somehow check what version of libobs a plugin is built against - and if it's old enough, ignore the notation rule but warn the user that the plugin may break/conflict.

Xaymar:
> @WizardCM: I feel a migration path / backwards compatibility would need to be fleshed out extensively.

I have added a possible migration path that should be compatible with old and new plugins.

> For example, somehow check what version of libobs a plugin is built against - and if it's old enough, ignore the notation rule but warn the user that the plugin may break/conflict.

Added, though it is considered an optional component.

Xaymar:
Closing as requested.

PR #42, opened 2021-06-08
Add RFC: User-local and System-wide plugins

Xaymar:
### Description
Allow OBS users to install user-local as well as system-wide plugins on every officially supported platform.

### Motivation and Context
The requirement of administrator rights to add plugins confuses users for strange reasons.

### [Link to RFC](https://github.com/xaymar/obs-rfcs/blob/rfc42/text/0042-user-and-system-plugins.md)

### DIFF
--- a/text/0042-user-and-system-plugins.md
+++ b/text/0042-user-and-system-plugins.md
@@ -1,0 +1,26 @@
+# Summary
+Allow users to install and load plugins without requiring administrator privileges.
+
+# Motivation
+The requirement of administrator rights to add plugins confuses users for strange reasons.
+
+# Drawbacks
+- Introduces a security vulnerability on Windows, where it is common to run OBS as adminsitrator for the improved GPU scheduling.
+
+# Design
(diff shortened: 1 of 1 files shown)


Xaymar:
Split from #34.

tytan652:
About the "Load Priority" on Linux, user-local install should take priority before system-wide install.
This is how many things in Linux behave like desktop entries, application configuration.

Xaymar:
I picked system-wide ahead of user-local due to obs internal plugins being installed system-wide. Perhaps a third category of plugin loading directories may be necessary, `program-bundled`.

tytan652:
I was mainly thinking about a scenario where the user made the mistake (or not) to install system-wide and user-local the same plugin.
OBS should load the user-local rather than the system-wide one.

Xaymar:
Good point, I didn't actually consider such a case.

PatTheMav:
Weirdly enough this is a functionality that macOS already supports but which I want to remove for that platform, because system-wide installation is an anti-pattern (unnecessarily requiring elevated privileges and installer packages) that should be discouraged and installing plugins in the user's library should be the _only_ valid location.

The only exception should be plugins used by other processes (e.g. our DAL plugin) which are truly "system-wide".

Xaymar:
Closing as requested.

PR #43, opened 2021-06-12
OBS WebRTC Output, Input and Plugin APIs

DDRBoxman:
# Description

Expose APIs to service plugins that allow configuring WebRTC as an output transport.

# Motivation and Context

Services such as Millcast, Wowza, Janus, and Caffeine allow webrtc ingest. It would be a better user experience if users could use webrtc as a transport without using a fork.

### [Link to RFC](https://github.com/ddrboxman/rfcs/blob/patch-1/text/0043-webrtc-output.md)


Fenrirthviti:
Since this is already shipping with OBS, and it's my miss on getting this merged, going to merge now. This was high level enough that the current implementation meets these requirements.

Skipping FCP and going to merge.

PR #46, opened 2021-11-24
Add RFC: Allow plugins to access multiview

cg2121:
### Description
Allow plugins, such as NDI and Decklink output to use the multiview as an output.

### Motivation and Context
* Currently, the only way to output the multiview, is via the projector output. Users might want to use other ways to display the multiview.
* It would also allow each multiview to have its own settings, such as layout and sources

### Link to RFC:
https://github.com/cg2121/rfcs/blob/multiview-access/text/0046-multiview-access.md

### DIFF
--- a/text/0046-multiview-access.md
+++ b/text/0046-multiview-access.md
@@ -1,0 +1,53 @@
+- Start Date: 2021-11-24
+- RFC PR: #46
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


cg2121:
Closing, because of https://github.com/obsproject/obs-studio/commit/50372164d09664f6444a34e82f678820b3b145cc

PR #47, opened 2022-03-30
Add RFC: Switch to Advanced Docking System

tytan652:
### Summary

- Switch main dock system to [ADS](https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System)
- Make ADS dock state per profile
- Allow the user to switch between pre-made layouts
  - Plugins could register their own
  - Users could save their own
- Dock added by plugins with the old methods are added as "*Legacy dock*"

Note: The "central widget" design is kept for now.

### Motivation and Context

- Provide a better dock system to end-users.
- I consider that I made enough progress on my WIP branches to make this RFC.
- Also because I require feedback about the dock separation part.

- **Link to RFC:** https://github.com/tytan652/rfcs/blob/advanced_docking/text/0047-switch-to-advanced-docking-system.md
- **OBS with ADS test binaries**: https://github.com/tytan652/obs-studio/pull/6

My WIP branches:
- [Dock separation](https://github.com/tytan652/obs-studio/tree/dock_separation
)
- [Central widget separation](https://github.com/tytan652/obs-studio/tree/central_widget_separation
)
- [Switch to ADS](https://github.com/tytan652/obs-studio/tree/advanced_docking)
  - [CI builds](https://github.com/tytan652/obs-studio/pull/6)
- [Frontend API to add ADS dock](https://github.com/tytan652/obs-studio/tree/ads_frontend_api)
- [DockState per profile](https://github.com/tytan652/obs-studio/tree/per_profile_dockstate_no_migration)

Potential ideas suggestions completed by this RFC:
- https://ideas.obsproject.com/posts/1257/ui-layout-profiles
- https://ideas.obsproject.com/posts/418/docks-visible-when-obs-minimized
- https://ideas.obsproject.com/posts/615/ui-collections
- https://ideas.obsproject.com/posts/164/allow-saving-of-visible-docks-layouts


RytoEX:
> Wayland is not properly supported by Qt yet. If you use Wayland, then you should set the session type to x11: XDG_SESSION_TYPE=x11 ./AdvancedDockingSystemDemo. You will find more details about this in issue https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System/issues/288.

@GeorgesStavracas ?

tytan652:
> Wayland is not properly supported by Qt yet. If you use Wayland, then you should set the session type to x11: XDG_SESSION_TYPE=x11 ./AdvancedDockingSystemDemo. You will find more details about this in issue [githubuser0xFFFF/Qt-Advanced-Docking-System#288](https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System/issues/288).

I want to precise this is the same for Qt Docking System, it's not properly supported either.

Edit note: I wish to avoid having to add a platform like `OBS_NIX_PLATFORM_XWAYLAND` because both docking system don't work properly under Wayland because Qt maybe don't care about that or consider it not a priority at all.

GeorgesStavracas:
> > Wayland is not properly supported by Qt yet. If you use Wayland, then you should set the session type to x11: XDG_SESSION_TYPE=x11 ./AdvancedDockingSystemDemo. You will find more details about this in issue [githubuser0xFFFF/Qt-Advanced-Docking-System#288](https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System/issues/288).
> 
> @GeorgesStavracas ?

I feel somewhat hopeless that we'll ever see good Wayland support for Qt. It's not only qt-wayland still has glaring bugs; the commercial nature of Qt means that if we contribute with bugfixes, they take way too long to be reviewed, and even when they're accepted, they're not backported. The Wayland backend is third-class on their priorities, the core of Qt is itself modelled against X11 and Windows APIs, and qt-wayland is merely an afterthought, often incompatible with the abstractions in place. In other words, qt-wayland seems to be there to check some boxes for customers, not to have actual Wayland support.

So yeah, I don't think we'll see proper docking using subsurfaces and the xdg-positioner protocol any time soon. And even if Qt magically gains such support, ADS would still need to be adapted, which would probably mean another year or more.

(Though, I must say, I'm a little skeptical about the entire docking concept. Docking leaks app-specific logic into the window management realm, and then we have problems with tiling, window ownership, modality, and basically break the app on anything that is not the traditional floating window model. But I digress.)

I'm not sure if the expectation is for me to fix that? If so, sorry, I have little motivation to contribute to qt-wayland upstream these days. I still have a bitter taste in my mouth for trying to fix that Wayland crash last year. :slightly_frowning_face: 

> I wish to avoid having to add a platform like OBS_NIX_PLATFORM_XWAYLAND

There is no Xwayland platform at all :)  There is either X11, or Wayland. If an app is running as X11, it should pick X11 code paths; if an app is running as a Wayland app, it should pick Wayland paths; there is no in-between there.

tytan652:
> So yeah, I don't think we'll see proper docking using subsurfaces and the xdg-positioner protocol any time soon. And even if Qt magically gains such support, ADS would still need to be adapted, which would probably mean another year or more.

Just want to precise that ADS does not rely on Qt Docking System implementation. So if possible (I have no experience with Wayland development), a support of Wayland could be added to ADS.

GeorgesStavracas:
> Just want to precise that ADS does not rely on Qt Docking System implementation. So if possible (I have no experience with Wayland development), a support of Wayland could be added to ADS.

That's correct, what I was trying to say is that both Qt's docking system and ADS rely on lower level interfacing with the windowing system, which in this case doesn't exist for qt-wayland.

tytan652:
- Updated transition widget
- Added what could be done if the switch to ADS is done with the switch to Qt6
- Rename every function with `_adv_dock` to `_dock` or `_dock_2`

tytan652:
~~Put as draft to prioritize #53 over this.~~

Warchamp7:
Putting aside the Wayland issues, is there other comments or concerns here?

text/0047-switch-to-advanced-docking-system.md
```diff
@@ -0,0 +1,302 @@
...
+[![Native titlebar under GNOME](./0047-switch-to-advanced-docking-system/linux_native_floating_dock_titlebar.png)]()
+
+- QWidget based (WIP CSS)
+
+[![QWidget titlebar](./0047-switch-to-advanced-docking-system/linux_qwidget_floating_dock_titlebar.png)]()
+
+The QWidget one is used by default when using KWin based Desktop Environement like Plasma. And this titlebar requires some CSS in the themes to match it.
+
+So the QWidget based titlebar usage will be enforced thanks to a flag for the Dock Manager to "unify" experience between desktop environement.
+Theme makers may ignore to theme this bar.
```
WizardCM:
```suggestion
Theme authors won't be required to theme the titlebar.
```

text/0047-switch-to-advanced-docking-system.md
```diff
@@ -0,0 +1,302 @@
...
+
+### Dock state
+*The per profile dock state feature is not taken into account to describe those changes.*
+
+In the global config (`global.ini`).
+- `"dockState"` is kept for for backward compatiblity and no longer overwitten. It will be used if the following state is not present.
+- `"windowState"` is the state of the main window, since it does not store only the state of legacy docks.
```
WizardCM: I assume this stores both legacy dock position, and main window position?

text/0047-switch-to-advanced-docking-system.md
```diff
@@ -0,0 +1,302 @@
...
+### Reset UI action
+1. Legacy dock are hidden
+2. The default state written in XML is applied
+3. Each not shown OBSAdvDock/BrowserAdvDock dock have its position and size reseted.
+
+## Legacy dock
+
+**TODO: Talk about `obs_frontend_add_custom_qdock()`**
+
+The frontend API method `obs_frontend_add_dock()` is put in deprecation. And if dual switch only Qt5 and on Qt 6 the method is removed or does nothing, this is possible because plugin will need to be rebuilt agaisnt Qt6.
```
WizardCM: This needs updating based on current situation.

text/0047-switch-to-advanced-docking-system.md
```diff
@@ -0,0 +1,302 @@
...
+And dock added through this method are added to a sub-menu named `Legacy dock` of the Dock menu.
+
+[![Legacy dock menu](./0047-switch-to-advanced-docking-system/legacy_dock.png)]()
+
+When openning a legacy dock, a message will appear explaining that those docks will not meld wery well with "new" docks.
+
+The state of those are saved through `"windowState"` global config.
+
+## Per profile dock state
+Move the `"advDockState"` from global config to the profile and integration no longer store their own state. `"windowState"` is kept global.
```
WizardCM, line 157: I do worry slightly about dock layouts being tied to profiles, as I expect most users won't have a different layout for each profile. Do we maybe want to have a global one by default, and if a user creates a layout then tie it to the current profile by default?

tytan652: Unfortunately, this is also disambiguate the service integration that already does that into a more service-agnostic approach.
*It's not only profiles with integration enabled, it's every profiles.*

text/0047-switch-to-advanced-docking-system.md
tytan652: This is why I added TODOs, I'm working on it.

text/0047-switch-to-advanced-docking-system.md
tytan652: The sentence is for showing that theme authors will ignore it because they don't develop the them to work on the Windows, macOS and Linux.

tytan652:
Undraft to avoid the "*it's a draft so no need to review for now*" situation even if there is still some TODO.

text/0047-switch-to-advanced-docking-system.md
tytan652: Changed the text.

PR #48, opened 2022-04-10
Add RFC: Introduce Update Channels to OBS updater

derrod:
# Summary

Changes to add the option for Windows users to opt into branches containing unstable/beta releases.

# Motivation

Currently Betas/RCs only reach a small number of users, and require manual download and installation (outside of Steam).

Additionally, we may also occasionally provide branches to test specific fixes without releasing a full Beta/RC or manually sending builds to users.

### [View RFC](https://github.com/derrod/rfcs/blob/updater-updates/text/0048-updater-branches.md)


WizardCM:
Overall, I'm happy with everything in this RFC.

If a user switches to a branch such as beta, and then back to stable, will they be offered the latest stable build (which may be older than the beta build), or will they be kept on their current beta build until the next stable release is out?

WizardCM:
Had another thought - your screenshot only shows a predefined dropdown. Are you thinking of an editable combobox for custom branches that aren't entirely public?

derrod:
> If a user switches to a branch such as beta, and then back to stable, will they be offered the latest stable build (which may be older than the beta build), or will they be kept on their current beta build until the next stable release is out?

If you opt-out of a beta you'd be asked to run the updater to get back to the release branch (basically exactly how Steam branches work).

> Had another thought - your screenshot only shows a predefined dropdown. Are you thinking of an editable combobox for custom branches that aren't entirely public?

I don't know exactly how combo boxes work, perhaps that would be an option. Otherwise it could be relegated to config editing or a command line switch or something.

SuslikV:
Nightly builds (automated builds here at github actions, at least for Windows) not always running OK in comparison to official releases. Scripting (may be) not fully operational or else. Better to not provide nightly builds through UI for casual users.

I think that, the checkbox to enable auto-update not needed if update channels will be implemented. One channel can point to - "Never check for updates"

derrod:
> Nightly builds (automated builds here at github actions, at least for Windows) not always running OK in comparison to official releases. Scripting (may be) not fully operational or else. Better to not provide nightly builds through UI for casual users.

Fair, although we will use terminology to make it clear that nightly builds are "unstable".

> I think that, the checkbox to enable auto-update not needed if update channels will be implemented. One channel can point to - "Never check for updates"

No that's stupid. You might want to be opted into a channel without having automatic update checks enabled. For instance if you do select "nightly" you might not want to be notified every day about updates, but still receive nightly updates if you manually check for them.

derrod:
UI mockups for channel selection with Yami:
2022-08-02_00-53-30_MXkiOB
2022-08-02_00-53-26_EKiBwF

derrod:
Updated the RFC to reflect recent developments in the PR.

derrod:
Would still need some feedback from @notr1ch and @jp9000 to see if the proposed server-side changes (mostly related to changes in the directory structure for patches/files and requiring signatures for more files) are looking okay.

Pull request for macOS implementation is here: https://github.com/obsproject/obs-studio/pull/7723

Fenrirthviti:
Motioning for FCP, seconded by @RytoEX off-thread.

There is still a question regarding file changes, but we can address that as part of FCP. Jim has given his tentative approval (again off-thread).

PR #49, opened 2022-06-01
Add RFC: Multiple Video Mixes

chippydip:
Proposal to refactor libobs to support multiple video mixes which features like ISO recording and Virtual Cam from any source can be built on.

# Summary

Add the ability to render multiple "video mixes" where a "video mix" is ultimately a `render_texture` created by compositing scenes and sources that can be fed into one or more video encoders/outputs. 

# Motivation

OBS supports multiple video outputs allowing simultaneous streaming, recording, virtual camera, etc. These outputs, however, must all be derived from a single `render_texture`. This makes it impossible to support [Multiple Video Outputs (Selective Recording, ISO recording)](https://ideas.obsproject.com/posts/41/multiple-video-outputs-selective-recording-iso-recording) or to share just a webcam via virtual camera (to participate in a video conference while streaming) with the current architecture.

While plugins do exist to implement these features, they do so in a roundabout way. They provide a no-op filter which intercepts video from a source for the plugin to process. This completely bypasses the normal output pipeline of OBS.

### [View RFC](https://github.com/chippydip/obs-rfcs/blob/multiple-video-mixes/text/0049-multiple-video-mixes.md)


tt2468:
I'll start off by saying I definitely support the idea of supporting multiple AV pipelines within OBS. It's a feature that prevents many advanced users from being able to achieve what they really would like to using OBS. Indeed, I was even having a discussion with a good friend of mine yesterday about this very idea.

I think though that I should share a few concerns regarding the scope and objectives of such an addition. Please don't take this in a negative way, but adding such functionality to OBS in a way that is sustainable both for code maintenance and future expansions requires very extensive planning. There are a few notable cases where changes were made to OBS over its history that have led to unforeseen restrictions in later feature additions. If we do this at all, I say we need to do it "right"
- An ABI break seems necessary at the very least.
- Driving multiple pipelines using the same single render thread and GPU doesn't seem much better than the current method of using video_t objects to create side feeds. Multithreaded rendering should be heavily considered.
- Modularity should be a "requirement", with higher level intercommunication between pipelines a "goal". 3 render pipelines that all happen in one blob of logic would surely make maintenance and diagnostics a nightmare. This also goes towards the idea of threading things.

OBS was designed around the idea of one render pipeline. To change that is a lot of work no matter what approach is taken. That being said, is there enough value to this approach to justify the amount of work and future code maintenance? Perhaps making it easier to use multiple OBS processes at once may be worth trying to put more effort into. After all, if we could simply improve multi-instance functionality, then it would most definitely scale better horizontally. Some real world changes might look like:
- Texture sharing functionality allowing instances to share surfaces across processes and GPUs (pipe an input or scene into another OBS instance for further processing)
- Better control over the GPU selected for rendering by OBS
- Better output UI. Unified control and configuration for outputs
- Utilities and improvements for running multiple OBS instances (better config path management, maybe even launcher for multi-instance?)

chippydip:
Thanks for the feedback!

> There are a few notable cases where changes were made to OBS over its history that have led to unforeseen restrictions in later feature additions.

This is exactly the feedback I'm looking for from folks with more historical context than myself. As far as I can tell the additional channels in `obs_view` are basically unused (aside from audio devices) and the way the renderer currently treats them if they were (rendering on top of one another on the same `render_texture`) doesn't seem to be a useful feature. The core of this proposal is basically just to change that behavior to render each video channel in a view onto its own `render_texture` instead. If I've missed something about why this is a bad idea or how the current behavior is actually useful I'm hoping someone here can point that out.

> Driving multiple pipelines using the same single render thread and GPU doesn't seem much better than the current method of using video_t objects to create side feeds. Multithreaded rendering should be heavily considered.

As you say, building a truly multi-threaded renderer would be a much larger undertaking and would certainly require significantly more planning. I'd argue that it would essentially amount to a ground-up rewrite of `libobs` rather than being a refactor. Ultimately, though, I'm not sure if the value that would provide would be worth the effort. One of the most significant costs of rendering is simply moving data to and from the GPU and making the render multi-threaded doesn't really help with that part. As long as the render thread is mostly just making calls to the GPU and using proper sync points and such it's unlikely to be a bottleneck that would benefit from parallelization.

I think a multi-threaded renderer might make more sense in a multi-GPU system with one thread rendering on each GPU, but I think that would be a totally different RFC from what I'm proposing here.

> OBS was designed around the idea of one render pipeline. To change that is a lot of work no matter what approach is taken. That being said, is there enough value to this approach to justify the amount of work and future code maintenance?

This is what I'm trying to optimize for with this proposal. How can we maximize the benefits while minimizing costs (both up front and, more importantly, future maintenance)? I talked with @jp9000 about this initially and he mentioned taking a crack at it many years back and ending up with such a mess of a refactor that he just dropped the branch. With that in mind, I've tried to propose a change with as little impact as possible on the existing code while accomplishing the goal of being able to output multiple different video feeds.

I hope that we can find a way to not let perfect be the enemy of good. I think it's important to keep a long-term vision in mind and then find ways to work toward that vision one step at a time with small incremental changes. This proposal, I think, strikes the right balance between adding useful features that users want now, while also making sure not to paint the project into a corner that will prevent future progress toward that long-term vision.

chippydip:
Updated the doc with a link to the PR and posting here for visibility: https://github.com/obsproject/obs-studio/pull/6577

henke37:
How does this interact with features such as "scenes as sources" and down stream keyers?

chippydip:
This is the test code I was using to set my "Webcam" source as an aux output when starting the virtual camera:
image

Essentially you can just take any source/scene and set it via a call to `obs_set_output_source(source, channel)` and get the `video_t` for a channel with the new `obs_get_video_mix(channel)` function. Then just wire up encoders/outputs to that `video_t` object as you normally would and it all just works. Behind the scenes `libobs` keeps track of which `video_t` is for which channel so the rest of the existing APIs don't even need to specify a channel number explicitly.

Fenrirthviti:
Adding thoughts here from the UI perspective, on how we feel this should be implemented:

### Controls Dock (Settings icon next to vcam):
unknown

### Virtual Camera Dialog (opened from Settings icon):
unknown

### Output Types: 
- Internal
- Scene
- Source

### Output Selection: 
- Internal:
  - Program Output (Default)
  - Preview Output
- Scene (Sort alpha):
  - Scene 1
  - Scene 2
  - Scene 3
  - …
- Source (Sort alpha):
  - Source 1
  - Source 2
  - Source 3
  - …

chippydip:
Implementation PR has been updated to match the above mocks for Virtual Camera support: https://github.com/obsproject/obs-studio/pull/6577

Fenrirthviti:
Since all the additional content of this PR is sound and we agree on this approach, I'm going to enter final comment period with disposition merge.

PR #52, opened 2022-09-28
Add RFC: Create output manager

cg2121:
### Description
Add an output manager so that outputs can be managed in a single dialog.

### Motivation and Context
- Each output (NDI, Decklink, AJA, etc.) doesn't have to write their own UI, removing much deplicated code.
- It would make it easier to find bugs and fix them for all outputs instead of just one.

### Link
https://github.com/cg2121/rfcs/blob/output-manager-rfc/text/0052-output-manager.md

### DIFF
--- a/text/0052-output-manager.md
+++ b/text/0052-output-manager.md
@@ -1,0 +1,37 @@
+- Start Date: 2022-09-28
+- RFC PR: #52
+
+# Summary
+
+Add an output manager so that outputs can be managed in a single dialog.
+
+# Motivation
+
+- Each output (NDI, Decklink, AJA, etc.) doesn't have to write their own UI, removing much deplicated code.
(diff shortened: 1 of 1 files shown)


Fenrirthviti:
This seems like it would be more complex and extra steps for users to have to manage their non-stream/recording outputs. Many outputs are trying to offer different thing, so there isn't much (at least, as I see it) that would be really shared through this. Unless I'm misunderstanding, it would add an additional step where a user has to select Tools -> Output Manager -> NDI/Decklink/etc.

I think screenshots or mockups of what this would look like, and a better explanation on what code would be simplified would help here. If there's something lacking from OBS to creating these custom outputs properties, would it make more sense to handle this as purely a backend/plugin API problem instead of trying to bolt everything to an outputs manager?

Fenrirthviti:
As there has been no response or clarification on this, moving to FCP with disposition close.

PR #54, opened 2023-06-14
Add RFC: libobs keychain APIs

derrod:
### Summary

Proposes a libobs platform API for keychain access which can be used to securely store credentials.

### Motivation and Context

While the credentials stored by OBS in profiles are not of severe sensitivity due to limited scope, as a modern desktop application we should still do our best to store them securely.

## [Link to RFC](https://github.com/derrod/rfcs/blob/libobs-keychain/text/0000-libobs-keychain-api.md)

### DIFF
--- a/text/0000-libobs-keychain-api.md
+++ b/text/0000-libobs-keychain-api.md
@@ -1,0 +1,78 @@
+# Summary
+
+This RFC proposes the addition of OS keychain APIs to the libobs platform utilities to faciliate secure storage of credentials.
+
+# Motivation
+
+While the credentials stored by OBS in profiles are not of severe sensitivity due to limited scope, as a modern desktop application we should still do our best to store them securely.
+
+# Detailed Changes
+
+This RFC proposes the addition of the following libobs API methods:
+
+- `bool os_keychain_available(void)`
+- `bool os_keychain_save(const char *label, const char *key, const char *data)`
+- `bool os_keychain_load(const char *label, const char *key, char **data)`
+- `bool os_keychain_delete(const char *label, const char *key)`
+
+As the API requires OS-specific implementations it shall become part of the `util/platform.h` header.
+
+**Implementation notes:**
+
+Operating-system backends:
+- Windows: `wincred.h` / `Cred*` Win32 API functions
+- macOS: Keychain Services API
+- Linux: `libsecret`
+
+The `label` is a user-visible component that may be displayed in keychain management applications as the label or name of an entry. It must be specified, ASCII only, and must be human-readable.  
+In most cases it should identify the component and contents, e.g. for OAuth credentials stored by the OBS Studio UI this could be "OBS Studio OAuth Credentials".  
+An entry can only be loaded or deleted if the label and key match the ones specified when saving.
+
+`os_keychain_available()` will always return `true` on Windows and macOS, on Linux it will return `true` if libobs was compiled with libsecret support, but does not indicate whether a keychain backend is available.
+
+**Usage examples:**
+Saving:
+```c
+const char *label = "OBS Studio Stream Keys";
+const char *key = "StreamKey_abcef";
+const char *value = "mystreamkey";
+bool success = os_keychain_save(label, key, value);
+```
+
+Loading:
+```c
+const char *label = "OBS Studio Stream Keys";
+const char *key = "StreamKey_abcef";
+char *value = NULL;
+bool success = os_keychain_load(label, key, &value);
+
+... do stuff ...
+
+if (value)
+    bfree(value);
+```
+
+Deletion:
+```c
+const char *label = "OBS Studio Stream Keys";
+const char *key = "StreamKey_abcef";
+bool success = os_keychain_delete(label, key);
+```
+
+# Drawbacks
+
+Accessing keychain information can reduce portability of profiles by not storing information in config files. OBS Studio may opt out of using a keychain if it is running in portable mode.
+
+In some cases accessing the keychain may require user confirmation, the API may block in such cases, which has to be accounted for by API users.
+
+If no keychain is available API users will have to fall back to storing data otherwise (e.g. obs data or config files).
+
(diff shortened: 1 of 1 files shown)


text/0000-libobs-keychain-api.md
```diff
@@ -0,0 +1,69 @@
...
+bool success = os_keychain_delete(key);
+```
+
+# Drawbacks
+
+Accessing keychain information can reduce portability of profiles by not storing OAuth information or stream keys in plaintext files.
+
+In some cases accessing the keychain may require user confirmation, the API may block in such cases, which has to be accounted for by consumers.
+
+On Linux the situation may be more complicated. While both GNOME and KDE provide a "Secret Service", other desktop environments might not. This will require further investigation.
```
tytan652: - Not KDE but Plasma or KDE Plasma, the KDE name no longer relate directly to the desktop environment
- GNOME and Plasma indeed can provide their own implementation of `org.freedesktop.secrets` (gnome-keyring and kwallet) but those are not bound to the desktop environment (e.g. kwallet can be used on GNOME).
- KeepassXC is also a provider of `org.freedesktop.secrets`

derrod: So that would require detecting if the system has any provider for `org.freedesktop.secrets` available, and otherwise either disable any functionality requiring more sensitive credentials (for the purpose of this, I would consider stream keys to not be super sensitive), or fall back to an OBS implementation (e.g. plaintext file).

derrod:
In implemeting this in https://github.com/obsproject/obs-studio/pull/9122 I was wondering if perhaps there is a need for something like `bool os_keychain_available(void)` to query keychain support? This may be important mostly on Linux where the keychain is not guaranteed to be provided by the OS e.g. to be able to determine whether or not to fall back to unsecured file storage or warn the users.

derrod:
Updated with `os_keychain_available()` and a `label` parameter that is used to specify a user-facing name for the group/service a keychain entry belongs to, e.g. "OBS Studio OAuth Credentials".

text/0000-libobs-keychain-api.md
```diff
@@ -0,0 +1,78 @@
...
+Deletion:
+```c
+const char *label = "OBS Studio Stream Keys";
+const char *key = "StreamKey_abcef";
+bool success = os_keychain_delete(label, key);
+```
+
+# Drawbacks
+
+Accessing keychain information can reduce portability of profiles by not storing information in config files. OBS Studio may opt out of using a keychain if it is running in portable mode.
```
PatTheMav, line 64: On macOS (and maybe Windows 11?) keychain items can be cloud-based, so using the keychain would _actually_ increase portability vs the status quo.

text/0000-libobs-keychain-api.md
```diff
@@ -0,0 +1,78 @@
...
+
+**Implementation notes:**
+
+Operating-system backends:
+- Windows: `wincred.h` / `Cred*` Win32 API functions
+- macOS: Keychain Services API
+- Linux: `libsecret`
+
+The `label` is a user-visible component that may be displayed in keychain management applications as the label or name of an entry. It must be specified, ASCII only, and must be human-readable.  
+In most cases it should identify the component and contents, e.g. for OAuth credentials stored by the OBS Studio UI this could be "OBS Studio OAuth Credentials".
```
PatTheMav, line 28: For better discoverability (and transparency) I'd prefer if _any_ keychain item is automatically prefixed with "OBS Studio" followed by the label chosen by any given implementation.

This would reduce the need for implementors to be "good citizens" and prefixing their labels themselves.

text/0000-libobs-keychain-api.md
derrod: The intention of portable mode is to have a self-contained instance that can be kept on e.g. a USB thumb drive I don't think that using a keychain makes sense there. And to avoid synchronisation issues in general I think the data should be kept local either to a machine or portable instance.

As far as I can tell credentials stored using the Win32 APIs on Windows aren't synced either, and macOS doesn't support portable mode.

text/0000-libobs-keychain-api.md
derrod: I agree, though that gets into the libobs vs OBS Studio discussion. While I personally think that this distinction doesn't matter in any practical sense, others have disagreed. Perhaps this could be solved by setting a prefix to use on libobs initialisation or something.

PatTheMav: Right, that would require an API function that allows the frontend to set the prefix.

Is this functionality serving a need of `libobs` or of our UI? Because if it's the latter, code would only be implemented in the UI and as such the implementation can use the default "OBS Studio" prefix, because it's not a `libobs` functionality.

text/0000-libobs-keychain-api.md
PatTheMav: Right "portable" as in "stored with the application data", not "portable" as in "data follows me between machines".

Keep in mind that iCloud Keychain is the suggested "default" keychain on macOS - keeping application keychain data local is considered "deprecated".

text/0000-libobs-keychain-api.md
derrod: With the goal of migrating service integrations into plugins, I believed it would make the most sense to have this functionality within libobs rather than the frontend. Implementing it within the frontend API may also be possible, I'm not sure if there are any plugins that store credentials (e.g. websockets, cloud captions) that don't also link against the frontend.

PatTheMav: Yeah that makes sense then.

I wouldn't mind if the prefix ends up being `libobs`, though a "product name" would be common as a prefix and probably also more recognisable to users.

text/0000-libobs-keychain-api.md
derrod: The current implementation in PR obsproject/obs-studio#9122 just uses the default/iCloud keychain.

PR #58, opened 2023-08-12
Add RFC: Resource Usage APIs

derrod:
# Summary

This RFC proposes the addition of libobs and various object (sources, encoders, etc.) APIs that allow querying system resource usage.

# Motivation

When using larger scene collections with many sources it can be difficult to identify which elements use system resources.

For example:
- RAM and CPU usage of browser sources
- VRAM usage of image sources or delay filters
- CPU usage of x264 and other CPU encoders
- Network usage of a media source streaming in data

### [Rendered RFC](https://github.com/derrod/rfcs/blob/source-resource-usage/text/0058-resource-usage-apis.md)

### DIFF
--- a/text/0058-resource-usage-apis.md
+++ b/text/0058-resource-usage-apis.md
@@ -1,0 +1,98 @@
+# Summary
+
+This RFC proposes the addition of libobs and various object (sources, encoders, etc.) APIs that allow querying system resource usage.
+
+# Motivation
+
+When using larger scene collections with many sources it can be difficult to identify which elements use system resources.
+
+For example:
+- RAM and CPU usage of browser sources
+- VRAM usage of image sources or delay filters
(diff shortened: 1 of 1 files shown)


text/0000-resource-usage-apis.md
sebastian-s-beckmann, line 1: File number should match PR number. I also noticed this is wrong for `0000-fragmented-recording.md` as well (should have been `0051-fragmented-recording.md`), but that's already merged and I kinda don't want to create a PR as that would make RFC numbers skip one more, not sure how to proceed there.

derrod: Well you don't necessarily know the PR number before submitting a PR :P

Numbers are also not the only issue, we have two folders even though there should really be only one. That's something that could be cleaned up via PR or direct push.

PR #59, opened 2024-03-29
Add RFC: Sophisticated Multiviewer Configuration

MasterDalK:
### Description
This RFC is a proposal for an overhauled UI around managing multiviewers and their configuration in an effort to modernize and streamline the flow that a user would take to create a multiview, but also allowing multiple multiviewers like one would expect from other products.

### Motivation and Context
The current multiviewer configuration is relatively lacking compared to other switcher solutions in the broadcast space.

# [Link to RFC](https://github.com/MasterDalK/rfcs/blob/multiview_sophisticated_configuration/text/0059-multiview-sophisticated-configuration.md)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+desktop.ini
+*/desktop.ini
--- a/text/0059-multiview-sophisticated-configuration.md
+++ b/text/0059-multiview-sophisticated-configuration.md
@@ -1,0 +1,43 @@
+# Summary
+
+I would like to propose an overhaul on the way Multiviews are handled and configured in OBS. A lot of the ideas here are coming from combining what I believe to be a lot of the best cases of multiviwer configurations from a lot of big switcher brands (Grass Valley, Ross, Blackmagic Design, etc) as well as trying to think differently in that we are a fully digital compositing system. In that regard, I've also heavily researched the EVS dyvi system as its heavily gpu-based. The aspect ratio of each multiviewer box would be assumed to match the canvas resolution, as it works currently.
+
+![Here is my proposed example visual for how one would go about configuring a multiviewer.](https://i.imgur.com/I2S4GQ7.png)
+
+The Proposal:
+ * The control would be pulled out of the general settings tab and moved into its own settings tab to make it more free-form rather than having to conform to a dropdown.
+ * There would be tabs for each multiviewer head, with a plus similar to a browser to create a new multiviewer head.
+ * The layout would be built by adding rows first, with each row added prompting for how many columns would exist. For example if I want to add Preview and Program and nothing more, I would add a row then type in 2 columns to give me 2 half-width boxes.
+ * All content of the multiviewer would exist centred on the output, so if you only have 1 row of 2 things, it would simply appear letterboxed a quarter of the way down the screen.
+ * Boxes would be selectable, if a single box was selected a button would reveal itself to allow breaking out the box into 4 smaller boxes or unmerging that box. If multiple boxes were selected, you would only have a "Merge Boxes" button.
+
+![Breaking out Example](https://i.imgur.com/7Rdntzm.png) ![Merging example](https://i.imgur.com/7RZB8x0.png)
+ * The selection of what goes in each box would be handled via 2 dropdown menus. The first dropdown menu is populated with source types, and then the latter would filter all available sources by the source type, as Preview and Program are internal already, they could simply be denoted as "Internal" or "special" source type. This allows for quicker slotting in of not only scenes, but also regular sources meaning you could add a browser source, video capture device or game capture directly into the multiviewer.
+ * Tally for the borders would be done by checking the source being active in Program or being in preview, whereas Program and Preview would never light up of course. Scenes would be checked in a similar fashion. This is beneficial as that way if you have a scene in program that has multiple things in it (like a gameplay overlay) then all active sources would light up red instead of (in the current system) just your active scene.
+
+In terms of getting those outputs projected, I am unsure of having them in the View menu or from within each Multiviewer tab provides more merit one way or the other. I think either way is fine, as the actual set up of the Multiviewers is what I'm hoping to rework here. If they remain in the View menu, then can simply list each Multiviewer head and select display or Windowed out of the pop-up, if it exists on the settings modal then I imagine it'd be somewhat similar towards the bottom of the configuration pane, perhaps behind an "output" button. 
+
+However, for the outputs themselves: *the labels for each box* should not have a background on it, and the text should be translucent. Only the Blackmagic switchers have backgrounds on the labels and the backgrounds make it extremely difficult to see what's going on behind it albeit existing in a location where usually important information just inside of title & action safe stuff would go (such as a timer for a countdown slate or most bottom-central UI elements in games). Almost all other major switcher lines make the labels translucent on top of video, or underneath/above the video box and I think we should avoid that as it makes the boxes less compact.
+
+# Motivation
+
+*What problem is this solving?*
+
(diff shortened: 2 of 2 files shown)


exeldro:
Why should it be part of OBS itself and not be an OBS plugin like [Durchblick](https://obsproject.com/forum/resources/durchblick.1484/) and [Source Dock](https://obsproject.com/forum/resources/source-dock.1317/) can provide you customizable multiviews?

Warchamp7:
> Why should it be part of OBS itself and not be an OBS plugin like [Durchblick](https://obsproject.com/forum/resources/durchblick.1484/) and [Source Dock](https://obsproject.com/forum/resources/source-dock.1317/) can provide you customizable multiviews?

Multiview is a feature of OBS, I don't see any issues with striving to improve it.

Warchamp7:
Noting down a few thoughts I have while I'm here

-----

There have been efforts to move non-global things out of the main Settings window. I think Multiviews make sense as their own dedicated window, which is an approach we're looking to take with both Scene Collections and Profiles in the future.

Matching multiviews to the canvas resolution as noted in the RFC makes the most sense. Mixing ratios within a multiview will overcomplicate both implementation and UI. However, one potential concern is that canvas resolution is currently associated with Profiles.

Following from the above, we'd need to determine where this data lives. The current multiview simply grabs the first `n` scenes that are flagged to appear in it. A proper customization system will be Scene Collection specific, which desyncs it from where video resolution lives (Profiles). There are ongoing discussions around moving video resolution to be Scene Collection data which would solve this.

While building custom layouts is nice, we should retain presets as options. Selecting a preset would automate a layout definition the same way building it manually would.

As far as placement goes, we should mimic the existing multiview system of simply using the order of entries to place them into spots left to right, top to bottom. This would allow easily switching between different layouts. If a user edits their layout or selects a preset with a lower slot total, we should retain the additional entries in case they increase the number of slots again, to avoid data loss.

I am not clear about what the merge/break boxes idea solves

-----

As Exeldro linked above, there is a plugin called Durchblick that offers a customizable multiview utilizing a grid system. While that approach works, it is a bit restrictive in terms of possible layouts compared to some hardware switchers. 

I've built a configurable multiview system as an interactable browser source, I'll leave some of my notes on suggestions/ideas here based off my experience creating that.

https://github.com/obsproject/rfcs/assets/1554753/716d265c-8c8f-4542-9cad-b8742f06e5ea

**Spec Thoughts**
A multiview `layout` consists of a `container` and other properties.
A `container` consists of an array of `slot`s and other `container`s. 
_Only the top-most container of a layout is allowed to nest `container`s_
A `slot` contains a source to render

UI visualization is going to be difficult, as Qt does not have any systems for maintaining aspect ratio of widgets specifically. We'd probably have to just calculate widget geometry sizes so we'd want to ensure it's done in a way that can rerun on window resize (Or we restrict the Multiview management window to a fixed size)

Each `slot` and `container` stores an aspect ratio factor, with the minimum value being equal to the number of items in a row*
_*Specifically whichever axis is longest, in case we want to support vertical_
This factor is relative to the parent's dimensions. The top-most container of a layout would always have a factor of 1.

For example, given a multiview with dimensions of 1920x1080 a slot with a ratio factor of 1 would be 1920 x 1080 and full frame.
- A layout with 2 items in one row would have a ratio factor of 2 for each slot, making each one 960 x 540.
- A layout with 2 on top and 4 on bottom would have 2 slots at 960 x 540, and 4 slots at 480 x 270.

A secondary scaling factor `shortFactor` is calculated by adding up 1 / row total for each row. If this value is above 1, we scale each slot additionally by 1 / `shortFactor`

- A layout with 2 on top, 3 below, and another 3 below, we'd exceed our short axis factor. The base sizes would be 960 x 540 and 633.6 x 356.4. Our short factor would be (1 / 2) + (1 / 3) + (1 / 3) = 1.166. We'd scale each item by 1 / 1.166 = 0.857.

MasterDalK:
> Why should it be part of OBS itself and not be an OBS plugin like [Durchblick](https://obsproject.com/forum/resources/durchblick.1484/) and [Source Dock](https://obsproject.com/forum/resources/source-dock.1317/) can provide you customizable multiviews?

Multiviewers already exist, and a lot of people use it. If the feature exists already natively I'd love to see improvements to it. Especially for the "population at large" plugins like those are a bit... out of their depth to get up & running.

>I think Multiviews make sense as their own dedicated window, which is an approach we're looking to take with both Scene Collections and Profiles in the future.

Moving it out into its own settings window is perfectly fine! So long as the configuration exists somewhere with more detail then it does currently.

>Following from the above, we'd need to determine where this data lives. The current multiview simply grabs the first n scenes that are flagged to appear in it. A proper customization system will be Scene Collection specific, which desyncs it from where video resolution lives (Profiles). There are ongoing discussions around moving video resolution to be Scene Collection data which would solve this.

I agree with both tying the data to scene collections, as well as moving canvas resolution to the scene collection data.

>While building custom layouts is nice, we should retain presets as options. Selecting a preset would automate a layout definition the same way building it manually would.

Some presets sound good, but I think by having customizable layouts you can lower the amount of presets required (i.e horizontal + vertical 2+8 and not much else preset wise off the top of my head but I will do more thinking on this). The Dyvi has no presets and is confusing to learn from the onset, so I understand that bit. However being locked to only presets like Blackmagic (or old OBS) is incredibly frustrating and from my understanding Grass Valley and Ross switchers tend to have 2-3 presets available for selection as you go into the multiviewer configuration under a "reset to defaults" option.

>As far as placement goes, we should mimic the existing multiview system of simply using the order of entries to place them into spots left to right, top to bottom. This would allow easily switching between different layouts. If a user edits their layout or selects a preset with a lower slot total, we should retain the additional entries in case they increase the number of slots again, to avoid data loss.

This is along the lines of what I meant by having some level of sane defaults, no real qualms but the ability to put things "out of order" will be very important for anyone using it at a more robust level.

>I am not clear about what the merge/break boxes idea solves

The intention here was to create a (potentially) more simplified method of what you had created in your browser source multiviewer, essentially operating like a spreadsheet. This is also where my **add row** idea comes from. The AR calculation was a bit confusing for some users at a glance, and this method actually comes _also_ from the new Multiviewer configuration that Blackmagic has implemented for the Constellation series. In Constellations you can break up and re-merge some blocks to either add more or less slots. This would allow for vertical boxes to be created more easily, as well as being able to do weird stuff like the ones where you have 2 half-width boxes on top, 1 half-width box on the bottom left and then 4 1/4-width boxes in the bottom right quadrant. Having done a lot of pre-prod in spreadsheets for my multiviewers so I know what I need to do when I get on-site, its way faster to do it like this than the way you had it in your browser source or in pretty much anyway I've ever had to build a multiviewer in any switcher, which is why I think there's a lot to learn from that paradigm. For all the rough-edges and complaints I've had about Blackmagic products and ATEM's, building some multiviewers on Constellations last week at PAX was incredibly simple with this merge/break paradigm.

MasterDalK:
I missed this in my earlier comment but restricting the multiview configuration window to a fixed size is standard across all other hardware switchers (likely for the reasons you stated Warchamp) as far as I remember.

mate1213:
Hi!
I've found this feature request which is connected to multiview.
Can we do this too, in the context of the reworking, and adding functionality to multiview?
https://ideas.obsproject.com/posts/1349/vu-metter-on-multiview

A lot of switchers, already supporting this kind of feature like BM Atem Studio series.

mate1213:
Hi!
Do we have some ETA, when can we see this feature in release?
Br

Fenrirthviti:
> Hi! Do we have some ETA, when can we see this feature in release? Br

This is an RFC, which means it's still an open discussion and the design is not even finalized yet. There is no expected timeframe to these changes being accepted, if they are even accepted at all.

Please ensure comments are kept on-topic in the future, we strongly discourage comments asking "when will this be added" or anything that isn't direct discussion of the RFC itself.

PR #63, opened 2024-06-20
Add RFC: Codec Identifiers

derrod:
### Description

This RFC intends to facilitate discussion to find a way to move away from storing codec identifiers as strings. This is to avoid string comparisons and remove some redundancies such as outputs defining their own codec enumerations.

### Motivation and Context

String comparisons are slow, error prone (case sensitivity, typos, etc.), and generally ugly. Most outputs convert from strings to internal enumerations anyway which could be made redundant by providing them in the first place.

[Link to RFC](https://github.com/derrod/rfcs/blob/no-codec-strings-attached/text/0000-codec-identifiers.md)

### DIFF
--- a/text/0000-codec-identifiers.md
+++ b/text/0000-codec-identifiers.md
@@ -1,0 +1,95 @@
+# Summary
+
+This RFC intends to facilitate discussion to find a way to move away from storing codec identifiers as strings. This is to avoid string comparisons and remove some redundancies such as outputs defining their own codec enumerations.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


text/0000-codec-identifiers.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+#### Advantages
+
+- Simple enum in libobs
+- No external dependencies
+
+#### Drawbacks
+
+- Requires libobs API changes to add new codecs
```
tt2468, line 87: Wouldn't all three solutions require libobs API changes?

tytan652: The third requires to modify the API at **each codec addition** unlike the two others that relies on a third-party enum or FourCC (not an enum but integers).

PR #64, opened 2024-08-07
Add RFC: Remove Dynamic Audio Buffering

tt2468:
### Description
This RFC describes the desire to remove the dynamic audio buffering system from OBS.

### Motivation and Context
Described in RFC


Sean-Der:
This would be great. I commonly get 'Why is my WebRTC feed so slow'. The dynamic audio buffering is almost always the culprit.

I believe by default Twitch has been advising users to turn off dynamic audio buffering for WebRTC users. Need to find the doc

PR #66, opened 2026-04-08
Add RFC: Make audio monitoring a output plugin

cg2121:
## Summary
Move audio monitoring out of libobs and make it an output plugin.

## Motivation
Currently, when a source is being monitored, a new audio monitor is created for each source. The goal here is to have
a single mix where all of the sources are output to, thus avoiding sync issues when having multiple sources being
monitored.

[Link to RFC](https://github.com/obsproject/rfcs/blob/4283bd38735d8fd53e0e6d7d93b98c6f1b971461/text/0066-audio-monitoring-output.md)


pkviet:
Check the ASIO PR where I add monitoring tracks which output mixed monitored audio to the ASIO output plugin.

